PR #5, opened 2016-10-14
Feature/websocket trigger

deanrobertcook:
WIP still need to (re) write tests for this stuff, please **don't** merge

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/push/PushService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/push/PushService.scala
@@ -52,7 +52,6 @@ class PushService(context: Context, keyValue: KeyValueStorage, client: EventsCli
     //      maybe this would be enough for some conversations and we would not need to sync so much
     processHistoryEvents(notifications.notifications, new Date)
 
-    gcmService.notificationsToProcess ! false //Finished retrieving notifications, allow websocket to close
     if (notifications.lastIdWasFound) debug("got missing notifications, great")
     else {
       info(s"server couldn't provide all missing notifications, will schedule slow sync, after processing available events")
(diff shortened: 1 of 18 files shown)


zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -157,96 +152,50 @@ class PushServiceSignals {
...
-      verbose(s"onConnected() in state: $state")
-      state = Waiting
-      freshNotificationId = None
-    } else {
-      verbose(s"onDisconnected() in state: $state")
-      fetchLast.cancel()
-      state = Disconnected
-    }
-  }
+  signals.pushConnected.filter(_ == false).on(dispatcher)(_ => fetchLast.cancel())
```
zbsz: `signals.pushConnected.filter(_ == false)` this will generate a `Signal[Boolean]` which will always be empty or have value `false`, meaning it will fire change event only on first time `pushConnected` goes to false.
If `pushConnected` switches to true and false again, then resulting signal will not notify listeners, its value will always be false. I don't think this is what you really need, you could do this kind of filtering on EventStream, but on Signal it is risky.

zbsz:
I'm not sure about ignoring GCM notifications. Using GCM just as a signal to fetch notifications is interesting and will work in general, but it also adds a bit of delay, GCM contains notification that can be used right away, pretty often that actually is all the info we need, maybe it would be worth using it.
On bad network connecting to websocket and fetching notifications can be slow, can also just fail. We could display whatever we got from GCM and process notifications in background.
It would also be faster to only fetch `/notifications` without creating websocket connection when in background.

I'm also not sure about simplifying of notification logic, didn't go through all the code exactly, but there were two things that we wanted to ensure there:
- we want to process events in chronological order, meaning we want to first process old events from `/notifications` and only then process incoming events from websocket, some state changes rely on event ordering, if we process them in different order we may get some weird races
- last notification id should keep the id of last successfully processed notification, we need to make sure that we don't advance it too much, especially if processing fails somehow, if for example we update it to freshly received event, but fail to process old notifications which were being fetched at the same time from `/notifications`, this can lead to lost messages

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
deanrobertcook: Hmm, yeah I wanted to convert that into a `onChanged` stream first... thanks!

deanrobertcook:
Yeah, I was thinking it was a bit wasteful, and as step 2 I'll do what Marco suggested and filter out the messages that we've received via GCM, but still ensure that we check everything from `/notifications`. As a hotfix I'd say a bit of delay shouldn't be a problem. 

Now I think about it, I guess that turning on the websocket is kind of pointless if we fetch from `/notifications`. 

And okay, I'll check my logic again to ensure that I keep the last two points you mentioned.

PR #6, opened 2016-10-17
Separate Threading, Signals and other utils for modular export

deanrobertcook:
Bit of a fun exercise to pull signals and the necessary utils classes into a separate module so that we can publish them separately in bintray and create smaller learning projects with them, for those people new to SE and signals.

**UI project doesn't compile, please don't merge**


deanrobertcook:
Never mind, I think I'll do something else here...

PR #16, opened 2016-11-09
Refactor assets

deanrobertcook:
### WIP

Some big architectural changes here: 
* There is now just one `AssetData` class instead of `ImageAssetData` and `AnyAssetData`. Any differences in data between the types of asset data are represented by the `AssetMetaData` type.
* The asset data type contains optional remote information to make reasoning about it's sync state with the backend a bit easier
* Generic Content methods have been simplified to only accept and return this asset data class. I'm trying to ensure that for a given asset in storage, it's local and remote ids, as well as any already-set data are immutable. 
* Since we only ever cared about full/preview versions of an image asset, the asset type now contains just one reference to another asset as it's `previewId`. This feels a bit ugly, especially in dealing with giphy/user profile pictures, so I think I might move this to a `Seq` of `AssetIds` in the `ImageMetaData` type. Video assets can also then have just one `previewId` in their meta data type. This makes more sense since audio messages don't really have a preview.

Also - tests are borked


zmessaging/src/main/scala/com/waz/cache/CacheEntryData.scala
```diff
@@ -26,25 +26,24 @@ import android.database.sqlite.SQLiteDatabase
...
-  data: Option[Array[Byte]] = None,
-  lastUsed: Long = currentTimeMillis(),
-  timeout: Long = CacheService.DefaultExpiryTime.toMillis,
-  path: Option[File] = None,
-  encKey: Option[AESKey] = None,
-  fileName: Option[String] = None,
-  mimeType: Mime = Mime.Unknown,
-  fileId: Uid = Uid(),
-  length: Option[Long] = None)
+case class CacheEntryData(key: AssetId,
```
wire-android: Not sure about using `AssetId` here, cache service is very generic, it can store any arbitrary data, using asset id introduces weird dependency (cache -> assets) and signifies that cache can be only used for assets.

zmessaging/src/main/scala/com/waz/content/AssetsStorage.scala
```diff
@@ -18,35 +18,69 @@
...
 
 class AssetsStorage(context: Context, storage: Database) extends CachedStorage[AssetId, AssetData](new TrimmingLruCache(context, Fixed(100)), storage)(AssetDataDao, "AssetsStorage") {
-  import Threading.Implicits.Background
+  private implicit val dispatcher = new SerialDispatchQueue(name = "AssetsStorage")
+  import EventContext.Implicits.global
 
-  val onUploadFailed = EventStream[AssetData]()
+  private val remoteMap = new mutable.HashMap[RAssetId, AssetId]()
+  private val uriMap = new mutable.HashMap[Uri, AssetId]()
+  private val assetsById = new mutable.HashMap[AssetId, AssetData]()
```
wire-android: Are you storing all assets in memory now? 
Reading whole table on app start?
Why?
That could be a lot of data, it could use lots of memory and take lots of time.
Don't do that.

We can sometimes prefetch things this way, and maintain in memory index, but only when we are sure that it won't be too big, for sure we don't want to keep everything in memory.

zmessaging/src/main/scala/com/waz/content/WireContentProvider.scala
```diff
@@ -98,17 +98,18 @@ class WireContentProvider extends ContentProvider {
     }
   }
 
-  private def getDecryptedEntry(key: String) =
+  private def getDecryptedEntry(key: AssetId) =
     cache.getEntry(key) flatMap {
       case Some(entry) if entry.data.encKey.isDefined =>
-        cache.getOrElse(key + "#_decr_", cache.addStream(Uid().str, entry.inputStream, entry.data.mimeType, entry.data.fileName, Some(cache.intCacheDir))(Expiration.in(12.hours))) map (Some(_))
+        //TODO Dean: do we need to have different keys for encrypted/unencrypted cache entries?
+        cache.getOrElse(key, cache.addStream(key, entry.inputStream, entry.data.mimeType, entry.data.fileName, Some(cache.intCacheDir))(Expiration.in(12.hours))) map (Some(_))
```
wire-android: Yes, we need different keys. Just take a look what happens here now, you just found an entry in cache with `key`, and now you are calling `cache.getOrElse` with the same `key`, so it will always return the same data, and we end up with encrypted garbage.

The point of this is that cache service stores most of the data on external storage (sd card), this storage is unsafe, so we encrypt all data kept in there, later we need to decrypt that when we want to access it. Most of the time we can decrypt on the fly (when accessing using InputStream), but sometimes we actually need decrypted file, then we will decrypt it and store in internal storage, and we also want this new file to be cleared at some point, that's why we add it to cache.
Decrypted file is stored with small timeout, we don't want to keep it for long, because we don't want to use too much space in internal storage, that's also why we don't ant to replace encrypted entry with this one, it most likely has longer timeout, so we want it to stay on external storage.

zmessaging/src/main/scala/com/waz/model/AssetData.scala
```diff
@@ -20,307 +20,283 @@ package com.waz.model
...
+                     otrKey:      Option[AESKey]        = None,
+                     sha:         Option[Sha256]        = None,
+                     name:        Option[String]        = None,
+                     previewId:   Option[AssetId]       = None,
+                     metaData:    Option[AssetMetaData] = None,
+                     source:      Option[Uri]           = None,
+                     proxyPath:   Option[String]        = None,
+                     //TODO remove v2 attributes when transition period is over
+                     convId:      Option[RConvId]       = None, //TODO remove
+                     data64:      Option[String]        = None //TODO remove data from asset
```
wire-android: Planning to remote `data`? 
Are we dropping inline assets with v3?
We used inline data for previews, to make sure that preview is available without a need for separate download, not sure how v3 fits with that, but I would think that we will sometimes want this behaviour.

zmessaging/src/main/scala/com/waz/model/AssetData.scala
```diff
@@ -20,307 +20,283 @@ package com.waz.model
...
+  case class RemoteData(remoteId: Option[RAssetId]    = None,
+                        token:    Option[AssetToken]  = None,
+                        otrKey:   Option[AESKey]      = None,
+                        sha256:   Option[Sha256]      = None
+                       )
 
-  implicit lazy val ImageDataOrdering: Ordering[ImageData] = new Ordering[ImageData] {
-    import com.waz.model.ImageData.Tag._
+  //needs to be def to create new id each time
+  def NewImageAsset(id: AssetId = AssetId()) = AssetData(id = id, metaData = Some(AssetMetaData.Image(Dim2(0, 0))))
```
wire-android: I would use lower case name for def.

zmessaging/src/main/scala/com/waz/model/AssetData.scala
```diff
@@ -20,307 +20,283 @@ package com.waz.model
...
-
-  implicit lazy val ImageDataDecoder: JsonDecoder[ImageData] = new JsonDecoder[ImageData] {
-    override def apply(implicit js: JSONObject): ImageData = {
-      val key = js.opt("otrKey") match {
-        case o: JSONObject => Try(implicitly[JsonDecoder[SignalingKey]].apply(o).encKey).toOption
-        case str: String => Some(AESKey(str))
-        case _ => None
+  object IsVideo {
+    def unapply(asset: AssetData): Boolean = {
+      asset match {
```
wire-android: `asset.metaData match {` ?

zmessaging/src/main/scala/com/waz/model/AssetData.scala
```diff
@@ -20,307 +20,283 @@ package com.waz.model
...
-      image.remoteId.foreach(id => o.put("remoteId", id.str))
-      image.data64.foreach(o.put("data", _))
-      o.put("sent", image.sent)
-      image.url.foreach(o.put("url", _))
-      image.proxyPath.foreach(o.put("proxyPath", _))
-      image.otrKey.foreach { k => o.put("otrKey", k.str) }
-      image.sha256.foreach { s => o.put("sha256", s.str) }
+  object IsAudio {
+    def unapply(asset: AssetData): Boolean = {
+      asset match {
```
wire-android: `asset.metaData match {` ?

zmessaging/src/main/scala/com/waz/model/AssetData.scala
```diff
@@ -20,307 +20,283 @@ package com.waz.model
...
   }
 
-  implicit lazy val ImageAssetDataEncoder: JsonEncoder[ImageAssetData] = new JsonEncoder[ImageAssetData] {
-    override def apply(data: ImageAssetData): JSONObject = JsonEncoder { o =>
-      o.put("id", data.id.str)
-      o.put("convId", data.convId.str)
-      o.put("versions", JsonEncoder.arr(data.versions)(ImageData.ImageDataEncoder))
+  object WithDimensions {
+    def unapply(asset: AssetData): Option[Dim2] = {
+      asset match {
```
wire-android: `asset.metaData match {`

zmessaging/src/main/scala/com/waz/model/AssetMetaData.scala
```diff
@@ -117,18 +121,18 @@ object AssetMetaData {
...
-      millis <- LoggedTry(bp.Duration.ofMillis(duration.toLong)).toOption
-    } yield Audio(millis)
+    def apply(context: Context, uri: Uri)(implicit ec: ExecutionContext): Future[Option[Audio]] =
+      for {
+        retriever <- MetaDataRetriever.get(context, uri)
+        Some(duration) <- Future.successful(Option(retriever.extractMetadata(METADATA_KEY_DURATION)))
+        Some(millis) <- Future.successful(LoggedTry(bp.Duration.ofMillis(duration.toLong)).toOption)
+        Some(mime) <- Future.successful(Option(retriever.extractMetadata(METADATA_KEY_MIMETYPE)))
+        loudness <- AudioLevels(context).createAudioOverview(uri, Mime(mime)).future
+      } yield Some(Audio(millis, loudness))
```
wire-android: Previous API was done this way to allow us to create asset in steps, so that we could send asset metadata even before full asset content is available, and only then load actual data, compute levels, etc, and upload updated asset version. Not sure how you do that now.

deanrobertcook:
@wire-android Yeah, sorry to break it to you ( :P ) but it was decided pretty quickly that upload placeholders would be taken out and we'd see how that works. 

And yeah, I've learnt a lot more about the cache now - I made some assumptions about it at the start which were wrong, so I'll move it back to the way it was and make the new assets work with it. 

And now with v3, everything (even previews) have to be uploaded and then a reference sent as (a part of) a message, so inlining data only makes sense for v2, which we're planning to stop soon. That's why it might be nicer to move the data field out of the asset data object and only load it when needed. This messes with creating image assets a bit though.

zmessaging/src/main/scala/com/waz/service/media/GiphyService.scala
```diff
@@ -17,48 +17,85 @@
...
+import scala.concurrent.Future
+
+class GiphyService(client: GiphyClient, assets: AssetService) {
   import Threading.Implicits.Background
-  
-  def getRandomGiphyImage: CancellableFuture[Seq[ImageAssetData]] = {
-    client.loadRandom().map {
-      case Nil => Nil
+
+  //TODO Dean - this is all a bit ugly - It might be better to re-introduce an Image AssetData container type that links previews together
```
wire-android: Maybe we don't need to care at all, are we ever displaying previews for assets sent with v2? 
Aren't those only images, and we no longer display polka dots, could we just ignore previews completely?
It seems to me that now previews only make sense for videos, we don't even need to send them for regular images, or am I missing something?

zmessaging/src/main/scala/com/waz/model/AssetData.scala
```diff
@@ -20,307 +20,289 @@ package com.waz.model
...
+    case (None, None, None, None) => Option.empty[RemoteData]
+    case _ => Some(RemoteData(remoteId, token, otrKey, sha))
+  }
 
-    override val idCol = Id
-    override val table = Table("Assets", Id, Asset, Data)
+  //TODO Dean: cache keys for images and cachekeys for other assets are handled a bit differently. Fix that
+  // We need the url as part of the cache key for images since they can be stored on the device and loaded over and over again
+  // The WireContentProvider, however, expects Audio messages NOT to be stored as a url in the cache, as it tries to pick the url
+  // apart and then find the cached entry based on the id. I should see if I can remove that.
```
deanrobertcook: The WireContentProvider currently inspects any URIs that match the pattern `content://com.waz.xys/{cache-key}`, extracts the cache key and then searches the cache based on that key. If the cache key is the full URI, then it won't find it... Currently causing me some headaches with audio...

zmessaging/src/main/scala/com/waz/service/MetaDataService.scala
```diff
@@ -28,6 +28,7 @@ import com.waz.utils.{LoggedTry, returning}
 
 import scala.util.Try
 
+//TODO Dean - we have two metadata services - can they be combined somehow?
```
deanrobertcook: Okay, didn't really look, just thought they might serve a similar purpose by their name :)

zmessaging/src/main/scala/com/waz/service/media/GiphyService.scala
deanrobertcook: Yeah, pretty much, but I thought perhaps we might want to re-introduce them later. 

I think the only thing might be with giphy (still/small vs full assets) and user profile pictures. But I guess even there I could ignore the preview.

zmessaging/src/main/scala/com/waz/service/assets/AssetLoader.scala
```diff
@@ -32,36 +33,35 @@ class AssetLoader(val context: Context, downloader: DownloaderService, assetDown
...
       case CachedAssetRequest(_, _, _) => CancellableFuture successful None
-      case LocalAssetRequest(cacheKey, uri, mime @ Mime.Video(), name) =>
+      case LocalAssetRequest(cacheKey, uri, mime@Mime.Video(), name) =>
         downloader.download(VideoAsset(cacheKey, uri, mime, name))(videoDownloader)
-      case LocalAssetRequest(cacheKey, uri, mime @ Mime.Audio.PCM, name) =>
+      case LocalAssetRequest(cacheKey, uri, mime@Mime.Audio.PCM, name) =>
         downloader.download(UnencodedAudioAsset(cacheKey, uri, name))(unencodedAudioDownloader)
       case LocalAssetRequest(cacheKey, uri, mime, name) =>
         downloader.download(AssetFromInputStream(cacheKey, () => AssetLoader.openStream(context, uri), mime, name))(streamDownloader)
-      case External(cacheKey, uri) if Option(uri.getScheme).forall(! _.startsWith("http")) =>
```
deanrobertcook: I do this check now where I create the load request in the asset data itself. I guess it might be safer to have it in both places, but didn't think it was necessary.

zmessaging/src/main/scala/com/waz/ui/MemoryImageCache.scala
```diff
@@ -23,55 +23,59 @@ import com.waz.ZLog._
...
    */
   private val lru = new TrimmingLruCache[Key, Entry](context, CacheSize(total => math.max(5 * 1024 * 1024, (total - 30 * 1024 * 1024) / 2))) {
-    override def sizeOf(key: Key, value: Entry): Int = value.size
+    override def sizeOf(id: Key, value: Entry): Int = value.size
   }
 
-  def get(id: AssetId, tag: String): Option[Bitmap] =
-    Option(lru.get(Key(id, tag))) flatMap {
-      case BitmapEntry(bitmap) => Some(bitmap)
+  def get(id: AssetId, req: BitmapRequest, imgWidth: Int): Option[Bitmap] =
```
wire-android: Is `BitmapRequest.width` the same as `imgWidth` here? Do we need separate parameter?

zmessaging/src/main/scala/com/waz/ui/Images.scala
```diff
@@ -22,79 +22,78 @@ import android.graphics.Bitmap
...
+  def createImageAssetFrom(bytes: Array[Byte]): api.ImageAsset = {
     if (bytes == null || bytes.isEmpty) ImageAsset.Empty
-    else {
-      val im = new ImageData("full", Mime.Unknown, 0, 0, 0, 0, 0, Some(RAssetDataId())) {
-        override lazy val data: Option[Array[Byte]] = Some(bytes)
-      }
-      returning(new LocalImageAsset(ImageAssetData(AssetId(), RConvId(), Seq(im))))(_.setMirrored(true))
-    }
+    //Created with camera, so don't cache since if the user cancels or sends, the image asset won't be needed again
+    new LocalImageAsset(AssetData.newImageAsset().copy(sizeInBytes = bytes.length, data = Some(bytes)))
```
wire-android: missing: `else`

zmessaging/src/main/scala/com/waz/ui/Images.scala
```diff
@@ -22,79 +22,78 @@ import android.graphics.Bitmap
...
+      verbose(s"fileImages: ${fileImages.sizeString}")
+      res
     }
   }
 
-  def getLocalImageAsset(data: ImageAssetData) = getOrUpdate(localImages)(data.id, new LocalImageAsset(data))
+  private def cacheImageAsset(id: AssetId, img: ImageAsset) = {
+    val res = getOrUpdate(images)(id, img)
+    verbose(s"images: ${images.sizeString}")
+    res
```
wire-android: You could use `returning` here:
```
returning(getOrUpdate(images)(id, img)) { _ => verbose(s"...") }
```

zmessaging/src/main/scala/com/waz/ui/MemoryImageCache.scala
deanrobertcook: `imgWidth` is the actual width of the original asset. From the old logic we would always want to fetch an bitmap that had a width "greater than either the request width or the original width". I guess because it doesn't make sense to try and create a bigger bitmap if the original image can't support it.

zmessaging/src/main/scala/com/waz/ui/Images.scala
deanrobertcook: yikes, good catch, thanks!

xuanzhaopeng:
V3 to V3

```
ImageAssetView com.waz.zclient.pages.extendedcursor.image.CursorGalleryItem{43011808 V.ED..C. ........ 362,375-549,560 #7f0f0236 app:id/iv__cursor_gallery_item} :: failed loading bitmap for LocalImageAsset(
                                                                     AssetData:
                                                                      id:            02982df2-8901-4513-bb5d-dfa5cd892e6b
                                                                      mime:          Mime()
                                                                      sizeInBytes:   0
                                                                      status:        UploadNotStarted
                                                                      rId:           None
                                                                      token:         None
                                                                      otrKey:        None
                                                                      sha:           None
                                                                      preview:       None
                                                                      metaData:      Some(Image(Dim2(0,0),))
                                                                      convId:        None
                                                                      data (length): 0
                                                                      other fields:  None, Some(file:///storage/emulated/0/c4fe9709.png), None, None
                                                                         ) id=02982df2-8901-4513-bb5d-dfa5cd892e6b
```

xuanzhaopeng:
Send a sketch from given V2+V3 build,  to a pure V2 build, I see a empty image container, I need to scroll up and down again, then I see this sketch.

deanrobertcook:
Okay, since QA is waiting on Actors and the ability to send in v3, I'm going to merge this. Tests are compiling again, but still quite broken. Will need to do some more work before making a release, if time permits...

PR #17, opened 2016-11-09
Removed the method addLocalSentMessage

JPonte:
The addLocalSentMessage method was being used only once and it was
messing with the order of the local messages. Using the addLocalMessage
with Status.SENT does the desired effect.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/messages/MessagesContentUpdater.scala
+++ b/zmessaging/src/main/scala/com/waz/service/messages/MessagesContentUpdater.scala
@@ -77,14 +77,6 @@ class MessagesContentUpdater(context: Context, val messagesStorage: MessagesStor
     } yield res
   }
 
-  def addLocalSentMessage(msg: MessageData) = Serialized.future("add local message", msg.convId) {
-    verbose(s"addLocalSentMessage: $msg")
-    lastSentEventTime(msg.convId) flatMap { t =>
-      verbose(s"adding local sent message to storage, $t")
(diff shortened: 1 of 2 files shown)


wire-android:
It messes with ordering on purpose, the idea was to make sure that e2ee verification message is put in correct place, messages ordering should tell user which messages were sent to unverified device and which didn't. Also in local storage it should never happen that there are some PENDING messages before a SENT one. Please verify a scenario where there are several unsent messages (slow network) and then we get OTR_UNVERIFIED when syncing the first of them. Do we want error message to be displayed below other unsent messages or above?

wire-android:
I also don't see how that addresses #242, the request was to notify on every identity change, we only notify verification state changes.

JPonte:
The problem is that it puts the "started using a new device" always before the last sent message. Even if the last sent message was sent while the conversation was still verified. If I select the "Send anyway" option the message gets put before the previous sent message and not the new one, because it goes before the last sent and not the last pending. And if there are messages from the other participant in the middle, it can get way up in the conversation and get invisible even though all these messages were still sent while the conversation is verified.

JPonte:
Right, I misunderstood the ticket.

wire-android:
I didn't want to say that `addLocalSentMessage` was right, seems that it is quite broken, what I want to say is that just using `addLocalMessage` instead may not be a complete solution.
From your description it seems that `lastSentEventTime` is broken, guess this method is not used anywhere else. Maybe you could fix it to make it work in every scenario. Or maybe we don't really need it, just take a look if previously mentioned scenario is handled now correctly/acceptable.

PR #20, opened 2016-11-17
[AN-4683] Username rules validation

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-4683

### DIFF
--- a/zmessaging/src/main/scala/com/waz/api/impl/Usernames.scala
+++ b/zmessaging/src/main/scala/com/waz/api/impl/Usernames.scala
@@ -18,21 +18,33 @@
 package com.waz.api.impl
 
 import com.waz.api
-import com.waz.api.{UsernameValidation, UsernamesRequestCallback}
+import com.waz.api.{UsernameValidation, UsernameValidationError, UsernamesRequestCallback}
 
 class Usernames extends api.Usernames{
   override def isUsernameAvailable(username: String, callback: UsernamesRequestCallback) = {
     //TODO: STUB
     callback.onUsernameRequestResult("", new UsernameValidation {
       override def isValid: Boolean = true
-      override def reason: String = ""
+      override def reason: UsernameValidationError = UsernameValidationError.NONE
     })
   }
   override def isUsernameValid(username: String): UsernameValidation = {
-    //TODO: STUB
-    new UsernameValidation {
-      override def isValid: Boolean = true
-      override def reason: String = ""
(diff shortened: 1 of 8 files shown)


zmessaging/src/main/scala/com/waz/api/impl/Usernames.scala
```diff
@@ -18,21 +18,33 @@
...
-      override def reason: String = ""
+      override def reason: UsernameValidationError = UsernameValidationError.NONE
     })
   }
   override def isUsernameValid(username: String): UsernameValidation = {
-    //TODO: STUB
-    new UsernameValidation {
-      override def isValid: Boolean = true
-      override def reason: String = ""
+    val MAX_LENGTH = 21
```
deanrobertcook: You'll want to move any static variables to a case object of `Usernames`, or else these get unnecessarily created with each instance of `Usernames`. Not a big deal here, but just something to be aware of

zmessaging/src/main/scala/com/waz/api/impl/Usernames.scala
```diff
@@ -18,21 +18,33 @@
...
+    }
+    username match {
+      case usernameRegex(_) => UsernameValidationImpl(isValid = true, UsernameValidationError.NONE)
+      case _ => UsernameValidationImpl(isValid = false, UsernameValidationError.INVALID_CHARACTERS)
     }
   }
 }
+
+case class UsernameValidationImpl (isValid: Boolean, reason: UsernameValidationError) extends UsernameValidation {
+}
```
deanrobertcook: Don't need `{}` at the end of the class definition.

Also, I think you could have a better name for the `UsernameValidationImpl`. However, I also don't think we'll ever need any other types of validation, so maybe we can just turn the trait itself directly into a case class.

zmessaging/src/main/scala/com/waz/api/impl/ZMessagingApi.scala
```diff
@@ -219,4 +219,6 @@ class ZMessagingApi(implicit val ui: UiModule) extends com.waz.api.ZMessagingApi
   override def getLogging: Logging = ui.cached(Uris.LoggingUri, new Logging)
 
   override def getConnectionIndicator = new ConnectionIndicator()
+
+  override def getUsernames() = new Usernames
```
deanrobertcook: don't need '()'

PR #21, opened 2016-11-18
Send giphy from Actor

xuanzhaopeng:


actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
```diff
@@ -285,6 +285,25 @@ class DeviceActor(val deviceName: String,
...
+              Successful
+            }
+          case _ =>
+            waitUntil(api.getGiphy.random())(_.isReady == true) map { results =>
+              conv.sendMessage(new Text("Via giphy.com"))
+              conv.sendMessage(new Image(results.head))
+              Successful
+            }
+        }
+        Successful
```
deanrobertcook: This will return `Successful` before the message is actually sent from the actor. Instead of matching an Option of the `searchQuery`, you should either `map` or `flatMap` the value to return a future that returns `Successful` when the message sending is complete. Tests might be a bit flaky otherwise.

PR #24, opened 2016-11-22
Bug fix: Cannot start timer for ephemeral image

xuanzhaopeng:
Timer should be started Not only for Asset, but also for ImageAsset.

(PS: the modification is on SE itself, not in Actor)


deanrobertcook:
nice catch

PR #25, opened 2016-12-01
Add sync request to check for availability of usernames

JPonte:
Uses sync handler instead of a direct request to avoid letting the UI handle with failed requests.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/HandlesService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/HandlesService.scala
@@ -1,0 +1,38 @@
+/*
+ * Wire
+ * Copyright (C) 2016 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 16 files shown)


zmessaging/src/main/scala/com/waz/service/HandlesService.scala
```diff
@@ -0,0 +1,39 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.service
+
+import com.waz.api.UsernameValidation
+import com.waz.model.Handle
+import com.waz.utils.events.{Signal, SourceSignal}
+
+class HandlesService {
+  val validatedHandles: SourceSignal[Map[Handle, UsernameValidation]] = Signal(Map())
```
deanrobertcook: Might be a bit nicer to write: `val validatedHandles = Signal(Map.empty[Handle, UsernameValidation])`. That way you don't have to worry about which implementation of Signal to use

zmessaging/src/main/scala/com/waz/service/HandlesService.scala
```diff
@@ -0,0 +1,39 @@
...
+  def updateValidatedHandles(validations: Seq[UsernameValidation]): Unit = {
+    validatedHandles mutate {original =>
+      val newHandles = (validations map (validation => Handle(validation.username) -> validation)).toMap
+      val result = original.filter(pair => !newHandles.contains(pair._1)) ++ newHandles
+      result
+    }
+  }
+
+  def getValidations(handles: Seq[Handle]) : Seq[UsernameValidation] = {
+    val currentHandles = validatedHandles.currentValue.getOrElse(Map())
```
deanrobertcook: generally, you want to avoid calling `currentValue` unless you really have to (normally you'd only do that in the UI). Since you have a signal, you can map it to a `Seq[UsernameValidation]` by doing:

```
val validations = validatedHandles.map { vhs =>
  vhs.filter(h => vhs.contains(h)).flatMap(h => vhs.get(h))
}
```

Or something like that. Then in the UI classes, you can have a loader that responds to the signal. I can explain that more in detail too.

PR #26, opened 2016-12-02
Add actor to support for updating unique user name

xuanzhaopeng:
PS: we call it user name in actor

### DIFF
--- a/actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
+++ b/actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
@@ -463,6 +463,17 @@ class DeviceActor(val deviceName: String,
         Successful
       }
 
+    case UpdateProfileUserName(userName) =>
+      val p = Promise[ActorMessage]()
+      waitUntil(api.getSelf)(_.getUser != null).map { self =>
+        self.setUsername(userName, new CredentialsUpdateListener {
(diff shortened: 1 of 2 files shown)


actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
```diff
@@ -463,6 +463,15 @@ class DeviceActor(val deviceName: String,
         Successful
       }
 
+    case UpdateProfileUserName(userName) =>
+      whenSelfLoaded { self =>
+        self.setUsername(userName, new CredentialsUpdateListener {
+          override def onUpdateFailed(code: Int, message: String, label: String): Unit = ()
```
deanrobertcook: You might want to return a failure message with the code and reason to help you debug, but up to you

actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
```diff
@@ -463,6 +463,15 @@ class DeviceActor(val deviceName: String,
...
+          override def onUpdated(): Unit = ()
```
deanrobertcook: Might also be better to return Successful from here, as it currently stands, Successful will be returned before the `setUsernames()` method completes, so it might not be ready in time before the tests continue

PR #36, opened 2016-12-15
Group search improvements

JPonte:
Groups can be found by the handles of their users.
Added indexes to the tables used in the group search to improve the performance of the query.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/ConversationData.scala
+++ b/zmessaging/src/main/scala/com/waz/model/ConversationData.scala
@@ -220,6 +220,11 @@ object ConversationData {
 
     import com.waz.model.ConversationData.ConversationType._
 
+    override def onCreate(db: SQLiteDatabase): Unit = {
+      super.onCreate(db)
+      db.execSQL(s"CREATE INDEX IF NOT EXISTS Conversation_search_key on Conversations (${SKey.name})")
+    }
+
     def establishedConversations(implicit db: SQLiteDatabase) = iterating(db.rawQuery(
       s"""SELECT *
          |  FROM ${table.name}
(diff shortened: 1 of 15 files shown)


zmessaging/src/main/scala/com/waz/model/ConversationData.scala
```diff
@@ -234,22 +239,36 @@ object ConversationData {
...
+         """.stripMargin
+      else
+        s"""SELECT DISTINCT c.*
+            |  FROM ${table.name} c, ${CM.table.name} cm, ${U.table.name} u
+            | WHERE cm.${CM.ConvId.name} = c.${Id.name}
+            |   AND cm.${CM.UserId.name} = u.${U.Id.name}
+            |   AND c.${ConvType.name} = ${ConvType(ConversationType.Group)}
+            |   AND c.${Hidden.name} = ${Hidden(false)}
+            |   AND u.${U.Id.name} != '${U.Id(self)}'
+            |   AND (c.${Cleared.name} < c.${LastEventTime.name} OR c.${Status.name} = ${ConversationStatus.Active.value})
```
deanrobertcook: Might be a bit nicer to group these two blocks together and use the if statement just to produce the last bit that's different. This will be a otherwise surely be a place where someone could introduce a bug later

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -69,11 +71,16 @@ class UserSearchService(queryCache: SearchQueryCacheStorage, commonConnsStorage:
         if ((cacheRefreshInterval elapsedSince cached.timestamp)) queryCache.getOrCreate(query, SearchQueryCache(query, Instant.now, None)).flatMap(_ => sync.syncSearchQuery(query)).logFailure()
         cached.entries.fold2(Signal.const(Vector.empty[UserData]), ids => usersStorage.listSignal(ids))
     }.map { users =>
+      verbose(s"")
```
deanrobertcook: Unnecessary log line

PR #39, opened 2017-01-06
Add the possibility of filtering the messages in a cursor

JPonte:
To be used on collections

### DIFF
--- a/zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
+++ b/zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
@@ -20,6 +20,7 @@ package com.waz.content
 import java.util.concurrent.ConcurrentHashMap
 
 import android.content.Context
+import android.support.v4.util.LruCache
 import com.waz.ZLog._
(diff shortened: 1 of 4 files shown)


zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
```diff
@@ -69,6 +69,9 @@ class MessagesStorage(context: Context, storage: ZmsDatabase, userId: UserId, co
       Future.successful
     }
 
+  def msgsFilteredIndex(conv: ConvId, messageFilter: MessageFilter): Future[ConvMessagesIndex] =
+    Future(new ConvMessagesIndex(conv, this, userId, users, convs, msgAndLikes, storage, filter = Some(messageFilter)))
```
JPonte: I thought about it but it would need a different structure since the key now would be the convId and the filter not just the convid. But as soon as we implemented filtering by content text for example, the map would get easily crowded. Maybe a LRU could be useful there.

It looks fishy indeed, I'll take a look.

PR #40, opened 2017-01-09
Actor: Fix send file for image asset

xuanzhaopeng:
Fix actor about to send image asset by File transfer.

Test with .txt , image file, and video file

### DIFF
--- a/actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
+++ b/actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
@@ -385,10 +385,15 @@ class DeviceActor(val deviceName: String,
         val file = new File(path)
         val assetId = AssetId()
         zmessaging.cache.addStream(CacheKey(assetId.str), new FileInputStream(file), Mime(mime)).flatMap { cacheEntry =>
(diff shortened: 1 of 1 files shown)


actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
```diff
@@ -385,10 +385,14 @@ class DeviceActor(val deviceName: String,
         val file = new File(path)
         val assetId = AssetId()
         zmessaging.cache.addStream(CacheKey(assetId.str), new FileInputStream(file), Mime(mime)).flatMap { cacheEntry =>
-          val asset = impl.AssetForUpload(assetId, Some(file.getName), Mime(mime), Some(file.length())) {
-            _ => new FileInputStream(file)
+          if(mime.startsWith("image/")) {
```
deanrobertcook: We also have the `Mime` class, which does this checking for you , so you can do:

```
Mime(mime) match {
  case Mime.Image() => // handle image case
  case _ => //handle other cases
}
```

But that's just a style point, not really necessary :)

deanrobertcook:
Merge when you're comfortable

PR #41, opened 2017-01-10
RemoteActor: Update JCE configuration JDK 8u102

xuanzhaopeng:
Ref : https://github.com/jruby/jruby/issues/4101

Change the JCE configuration


tests/utils/src/main/scala/com/waz/testutils/package.scala
```diff
@@ -298,6 +307,8 @@ object JCE {
       instance.setAccessible(true)
       defaultPolicy.add(instance.get(null).asInstanceOf[Permission])
     } catch {
+      case e: ClassNotFoundException =>
+        println(s"Unable to enable unlimited-strength crpto: $e")
```
mykola-mokhnach: -> crypto

PR #42, opened 2017-01-10
DeviceActor: Use whenConversationExist instead of withConv or getConv

xuanzhaopeng:
The reason is : withConv and getConv doesn't wait until the conversation exist. If I do SendFile after Login immediately,  and if I do getConv or withConv, the actor will be failed.

### DIFF
--- a/actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
+++ b/actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
@@ -242,16 +242,17 @@ class DeviceActor(val deviceName: String,
       }
 
     case GetMessages(rConvId) =>
-      for {
-        conv <- getConv(rConvId)
-        idx <- zmessaging.messagesStorage.msgsIndex(conv.id)
-        cursor <- idx.loadCursor
-      } yield {
+      whenConversationExistsFuture(rConvId) { conv =>
+        for {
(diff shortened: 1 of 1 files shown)


actors/base/src/main/scala/com/waz/provision/DeviceActor.scala
```diff
@@ -242,16 +242,17 @@ class DeviceActor(val deviceName: String,
...
-        conv <- getConv(rConvId)
-        idx <- zmessaging.messagesStorage.msgsIndex(conv.id)
-        cursor <- idx.loadCursor
-      } yield {
-
-        ConvMessages(Array.tabulate(cursor.size) { i =>
-          val m = cursor(i)
-          MessageInfo(m.message.id, m.message.msgType, m.message.time)
-        })
+      whenConversationExistsFuture(rConvId) { conv =>
```
mykola-mokhnach: does this future have a timeout?

xuanzhaopeng: Hi, just with help from Dean, we add 30 seconds as default timeout

mykola-mokhnach: Set it to 120  seconds- we have some very large conversations

xuanzhaopeng: Done 120 as default

PR #46, opened 2017-01-18
Method that gets the position of a message in the cursor

JPonte:


zmessaging/src/main/scala/com/waz/content/MessagesCursor.scala
```diff
@@ -179,6 +179,28 @@ class MessagesCursor(conv: ConvId, cursor: Cursor, override val lastReadIndex: I
...
+    val currentPos = cursor.getPosition
+    val pos = cursorBinarySearch(messageData) match {
+      case -1 => None
+      case i => Some(i)
+    }
+    cursor.moveToPosition(currentPos)
+    pos
+  }
+
+  private def cursorBinarySearch(messageData: MessageData, from: Int = 0, to: Int = cursor.getCount - 1): Int = {
```
wire-android: Binary search on db cursor is not necessarily a good idea. Cursor only keeps some window of elements, and binary search jumps around, each such jump could require full window reload from db (io access).

PR #50, opened 2017-01-24
[AN-4572] No internet bar fix

hakonbo:


zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
```diff
@@ -22,17 +22,21 @@ import android.net.{ConnectivityManager, NetworkInfo}
...
+import scala.concurrent.duration._
+
 class NetworkModeService(context: Context) {
   import NetworkModeService._
 
   private implicit val ev = EventContext.Global
 
   private lazy val connectivityManager = context.getSystemService(Context.CONNECTIVITY_SERVICE).asInstanceOf[ConnectivityManager]
   private lazy val telephonyManager = context.getSystemService(Context.TELEPHONY_SERVICE).asInstanceOf[TelephonyManager]
+  private val scheduledNetworkCheckTimeout = 15.seconds
```
deanrobertcook: Best to put these constant vals in a companion object (not so important here, since `NetworkModeService` is a singleton), but a good habit to get into

zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
```diff
@@ -41,8 +45,10 @@ class NetworkModeService(context: Context) {
   }
   context.registerReceiver(receiver, new IntentFilter(ConnectivityManager.CONNECTIVITY_ACTION))
   updateNetworkMode()
+  CancellableFuture.delayed(scheduledNetworkCheckTimeout)(updateNetworkMode(true))
 
-  def updateNetworkMode(): Unit = {
+  def updateNetworkMode(scheduled: Boolean = false): Unit = {
+    if (scheduled) CancellableFuture.delayed(scheduledNetworkCheckTimeout)(updateNetworkMode(true))
```
deanrobertcook: nice, that actually does away with the need for a `private var`

hakonbo:
There is no need to re-register the broadcast receiver. If the network conditions changes while app is running in foreground, we still get the call to `onReceive`, even after we have missed one in the background. You can verify this by turning wifi off/on next time you see the stuck no internet indicator. If we can't figure out why we are missing calls to `onReceive` when app is in the background, I think the only thing we can do is to periodically update `networkMode`. We can of course log that we missed an update, but not sure that is very useful.

zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
hakonbo: Sure, I'll move it, move it. I like to move it, move it.

PR #54, opened 2017-01-26
Re-register broadcast receiver

hakonbo:
Thoughts?


zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
```diff
@@ -43,6 +43,7 @@ class NetworkModeService(context: Context) {
   val receiver = new BroadcastReceiver {
     override def onReceive(context: Context, intent: Intent): Unit = updateNetworkMode()
   }
+  val ctx = context
```
deanrobertcook: you can just rename the variable if you don't like it :)

PR #58, opened 2017-02-02
Removed token refresh threshold period

hakonbo:


zmessaging/src/main/scala/com/waz/znet/AuthenticationManager.scala
```diff
@@ -73,48 +74,35 @@ class AuthenticationManager(client: LoginClient, user: CredentialsHandler) exten
...
+  def currentToken(): Future[Either[Status, Token]] = tokenPref() flatMap {
+    case Some(token) if !isExpired(token) => Future.successful(Right(token))
+    case _ => {
+      debug(s"currentToken refresh, \ncredentials: ${user.credentials}")
+      loginFuture = loginFuture.recover {
+        case ex =>
+          warn(s"login failed", ex)
+          Left(Cancelled)
+      } flatMap { _ =>
+        debug(s"No access token, or expired, cookie: ${user.cookie}")
```
deanrobertcook: We could print the expired token here

PR #61, opened 2017-02-09
Use same cache key for scaled images

hakonbo:
What do you think @deanrobertcook?

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/images/ImageAssetGenerator.scala
+++ b/zmessaging/src/main/scala/com/waz/service/images/ImageAssetGenerator.scala
@@ -24,7 +24,6 @@ import com.waz.bitmap.BitmapUtils.Mime
 import com.waz.bitmap.{BitmapDecoder, BitmapUtils}
 import com.waz.cache.{CacheEntry, CacheService, LocalData}
 import com.waz.model
-import com.waz.model.AssetMetaData.Image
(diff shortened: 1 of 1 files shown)


zmessaging/src/main/scala/com/waz/service/images/ImageAssetGenerator.scala
```diff
@@ -119,22 +118,22 @@ class ImageAssetGenerator(context: Context, cache: CacheService, loader: ImageLo
...
   }
 
   private def saveFormat(mime: String, forceLossy: Boolean) =
     if (!forceLossy && mime == Mime.Png) Bitmap.CompressFormat.PNG
     else Bitmap.CompressFormat.JPEG
 
-  private def saveImage(id: AssetId, image: Bitmap, mime: String, options: CompressionOptions): CancellableFuture[(CacheEntry, Metadata)] =
-    cache.createForFile(CacheKey.fromAssetId(id), cacheLocation = Some(saveDir)).flatMap(saveImage(_, image, mime, options)).lift
+  private def saveImage(key: CacheKey, image: Bitmap, mime: String, options: CompressionOptions): CancellableFuture[(CacheEntry, Metadata)] =
+    cache.createForFile(key, cacheLocation = Some(saveDir)).flatMap(saveImage(_, image, mime, options)).lift
```
deanrobertcook: yeah, that is actually a much smarter way of fixing the problem than changing the cache key. Nice!

PR #67, opened 2017-02-14
Initial text search

JPonte:


tests/unit/src/test/scala/com/waz/content/MessageIndexStorageSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+  lazy val conv = ConversationData(ConvId(), RConvId(), None, UserId(), ConversationType.Group)
+  lazy val zms = new MockZMessaging() {
+    convsContent.insertConversation(conv).futureValue
+  }
+
+  lazy val messages = zms.messagesStorage
+
+  feature("Text search") {
+
+    scenario("") {
```
deanrobertcook: awesome test

zmessaging/src/main/scala/com/waz/content/MessageIndexStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+    storage(MessageContentIndexDao.updateMessages(updated.filter(m => m._2.msgType == Message.Type.TEXT || m._2.msgType == Message.Type.TEXT_EMOJI_ONLY || m._2.msgType == Message.Type.RICH_MEDIA))(_))
+  }
+
+  messagesStorage.onDeleted{ removed =>
+    storage(MessageContentIndexDao.removeMessages(removed)(_))
+  }
+
+  def searchText(contentSearchQuery: ContentSearchQuery, convId: Option[ConvId]): Future[MessagesCursor] ={
+    convId match {
+      case Some(conv) => storage(MessageContentIndexDao.findContent(contentSearchQuery.toFtsQuery, convId)(_)).map(c => new MessagesCursor(conv, c, 0, Instant.now, loader))
```
deanrobertcook: make sure to match the None or _ case here if you expect an `Option[ConvId]`

zmessaging/src/main/scala/com/waz/content/MessageIndexStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  val finishedIndexing = Signal.future(false, updateOutdated())
+
+  messagesStorage.onAdded{ added =>
+    storage(MessageContentIndexDao.addMessages(added.filter(m => m.msgType == Message.Type.TEXT || m.msgType == Message.Type.TEXT_EMOJI_ONLY || m.msgType == Message.Type.RICH_MEDIA))(_))
+  }
+
+  messagesStorage.onUpdated{ updated =>
+    storage(MessageContentIndexDao.updateMessages(updated.filter(m => m._2.msgType == Message.Type.TEXT || m._2.msgType == Message.Type.TEXT_EMOJI_ONLY || m._2.msgType == Message.Type.RICH_MEDIA))(_))
```
deanrobertcook: you could take out the types of messages we want to be able to search into a `Set(TEXT, TEXT_EMOJI_ONLY, RICH_MEDIA`), and then use `contains` on the message type to avoid a bit of repetition here. Then save this logic in the companion object

zmessaging/src/main/scala/com/waz/model/MessageContentIndexDao.scala
```diff
@@ -0,0 +1,97 @@
...
+  val Content = text('content)(_.content)
+  val Time = timestamp('time)(_.time)
+
+  private lazy val transliteration = Locales.transliteration("Latin-ASCII; Lower")
+
+  override def onCreate(db: SQLiteDatabase) = {
+    db.execSQL(table.createFtsSql)
+  }
+  override val idCol = MessageId
+  override val table: Table[MessageContentIndexEntry] = Table("MessageContentIndex", MessageId, Conv, Content, Time)
```
deanrobertcook: can probably remove the type info here - keep things looking a bit tidier

zmessaging/src/main/scala/com/waz/model/MessageContentIndexDao.scala
```diff
@@ -0,0 +1,97 @@
...
+    deleteEvery(removed)
+  }
+
+  def updateOldMessages()(implicit db: SQLiteDatabase): Boolean ={
+    val queryString =
+      s"SELECT ${MessageDataDao.table.name}.* FROM ${MessageDataDao.table.name} " +
+        s"WHERE " +
+        s"(${MessageDataDao.table.name}.${MessageDataDao.Type.name} = '${MessageDataDao.Type(Message.Type.TEXT)}' " +
+        s"OR ${MessageDataDao.table.name}.${MessageDataDao.Type.name} = '${MessageDataDao.Type(Message.Type.TEXT_EMOJI_ONLY)}' " +
+        s"OR ${MessageDataDao.table.name}.${MessageDataDao.Type.name} = '${MessageDataDao.Type(Message.Type.RICH_MEDIA)}') " +
```
deanrobertcook: You can build this using the same set of searchable types I mentioned earlier and the `mkString` method on collections, something like 
```
searcheableTypes.map(t => s"${MessageDataDao.table.name}.${MessageDataDao.Type.name} = t")).mkString(" OR ")
```

That way if we add/remove types in the future, we won't forget to update this query and can save someone a headache

zmessaging/src/main/scala/com/waz/content/MessageIndexStorage.scala
```diff
@@ -0,0 +1,76 @@
...
+  private implicit val dispatcher = new SerialDispatchQueue(name = "MessageIndexStorage")
+
+  def updateOutdated(): CancellableFuture[Boolean] = {
+    storage{ MessageContentIndexDao.updateOldMessages()(_)}.flatMap{
+      case false => CancellableFuture.delay(UpdateOldMessagesThrottle).flatMap(_ => updateOutdated())
+      case _ =>  CancellableFuture.successful(true)
+    }
+  }
+
+  val finishedIndexing = Signal.future(false, updateOutdated())
```
deanrobertcook: I think for this use case, the `Signal#orElse` method is what you're looking for. It checks to see if a signal is empty and then applies some value, or else returns the value of the signal

PR #72, opened 2017-02-22
Fixed ascending check on cursor when the first messages have same timestamp

JPonte:


zmessaging/src/main/scala/com/waz/content/MessagesCursor.scala
```diff
@@ -202,6 +194,26 @@ class MessagesCursor(conv: ConvId, cursor: Cursor, override val lastReadIndex: I
...
+      val hasNext = cursor.moveToNext()
+      val second = Entry(cursor)
+      val comparison = first.time.compareTo(second.time)
+      if (hasNext && comparison == 0){
+        return checkNext()
+      }
+      if (comparison != 0) {
+        return comparison < 0
+      }
+      false
```
deanrobertcook: A little bit too java-ish. Generally returns are frowned upon, and a case statement or an if-else block makes things a bit clearer, especially for a method that is in essence purely functional. If someone comes along and refactors this, forgetting a `return` won't be caught by the compiler. So I think a better style is: 

```
if (hasNext && comparison == 0) checkNext 
else if (comparison != 0) comparison < 0
else false
```

Even better: methods like `isAscending` can be placed quite nicely in the companion object, since they are purely functional.

PR #80, opened 2017-03-06
Adding play.spotify.com to the list of known domains

makingthematrix:
Adding play.spotify.com to the list of known domains in order for [AN-5062] to work with Spotify links.

Changing how the known domains are recognized from the exact match to that it is sufficient for the domain name to be a substring.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/media/RichMediaContentParser.scala
+++ b/zmessaging/src/main/scala/com/waz/service/media/RichMediaContentParser.scala
@@ -25,6 +25,7 @@ import com.waz.ZLog._
 import com.waz.api.Message.Part
 import com.waz.api.Message.Part.Type._
 import com.waz.model.MessageContent
+import com.waz.sync.client.{SoundCloudClient, SpotifyClient, YouTubeClient}
 
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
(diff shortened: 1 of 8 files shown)


zmessaging/src/main/scala/com/waz/service/media/RichMediaContentParser.scala
```diff
@@ -53,11 +52,11 @@ class RichMediaContentParser {
       case _           => false
     }
 
-    def matchDomain(host: String): Part.Type = knownDomains.getOrElse(host, {
+    def matchDomain(host: String): Part.Type = knownDomains.find(entry => host.contains(entry._1)).getOrElse(host, {
```
wire-android: Isn't this too lenient? It will match links like `myownyoutube.com`. It's also slower as it always goes through whole map. 
Is there a reason to do it like that? Notice that `getOrElse` part handles subdomains.

makingthematrix: Right, it can also match links like myownyoutube.com . I'll revert it.

zmessaging/src/main/scala/com/waz/znet/ZNetClient.scala
```diff
@@ -245,8 +247,8 @@ object ZNetClient {
     case Response(ErrorStatus(), ErrorResponse(code, msg, label), _) =>
       warn(s"Error response to $name query: ${ErrorResponse(code, msg, label)}")
       Left(ErrorResponse(code, msg, label))
-    case resp @ Response(ErrorStatus(), _, _) =>
-      warn(s"$name query failed: $resp")
+    case resp @ Response(ErrorStatus(), body, headers) =>
+      warn(s"$name query failed: $resp, body: $body, headers: $headers")
```
deanrobertcook: I think the body and headers will get printed twice...

makingthematrix: OK, I'll revert it.

PR #81, opened 2017-03-06
Remove past participants

JPonte:


deanrobertcook:
You can just squash the commits into one

zmessaging/src/main/scala/com/waz/sync/client/ConversationsClient.scala
```diff
@@ -132,7 +132,7 @@ object ConversationsClient {
 
       ConversationData(
         ConvId(id.str), id, decodeOptString('name)(js) filterNot (_.isEmpty), decodeUserId('creator)(js), convType,
-        lastEventTime, 'status, decodeISOInstant('status_time),
+        lastEventTime, 'status,
```
deanrobertcook: Might be better to remove this reference to status and use some default value instead

deanrobertcook:
Do we also need the `muted_time` and `archived_time` attributes in conversation data? I see there are some code dependencies, but from a logical perspective, a conversation is either muted, archived, or it's not.

As discussed - saving the muted time is still a good idea if we ever want to have "timed" muting

PR #97, opened 2017-03-29
fixes ImageLoaderSpec, removes previews loading for bitmaps

makingthematrix:
Until now ImageLoader tried to download the same bitmap twice - once as a preview and once as the full image and if the full image was downloaded first, it cancelled the preview download. This commit removes previews download.
Also, it adds a test for animated gifs previews download and a small utils
class with one method 'whereAmI' which returns the stack trace as a string. It can be used in debug messages.


makingthematrix:
Please also take a look at my comment under AN-5072:
https://wearezeta.atlassian.net/browse/AN-5072?focusedCommentId=70788&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-70788

PR #102, opened 2017-04-04
[AN-5151] adds asset v3.1 receiving possibility

makingthematrix:
Uses generic-message-proto 1.19.0 which adds an optional enum, EncryptionAlgorithm, which can be either AES_CBC (the old one), or AES_GCM (the new one). 

The new algorithm is not implemented yet, so instead there is a new switch in PreferenceService, v31AssetsEnabled, by default set to false. Only if the switch was set to true and the AES_GCM was set in AssetData, the new algorithm would be used to decrypt received data.

Accidentally, the new generic-message-proto also incorporates one small change in Messages.Confirmation: messageId is changed to firstMessageId.


makingthematrix:
Build again.

PR #111, opened 2017-04-21
replaces references to SQLite with DB wrappers

makingthematrix:
Also fixes a few minor issues in AssetData:
* When Option[Loudness] was None, after serializing to json and deserializing it became Some[Loudness(empty vector)] - fixed
* 'equals' and 'hashCode' added to AssetStatus and URI, so now AssetData can be compared with another AssetData
* the name for the mime type field is 'mime' in the new version and 'mimeType' in the old one. Now both are handled correctly for the sake of testing (it seems it is not possible to mix them up in the production code anyway)


makingthematrix:
build again

PR #113, opened 2017-04-26
Download retries and Android-free unit tests.

makingthematrix:


zmessaging/src/main/scala/com/waz/service/images/ImageLoader.scala
```diff
@@ -251,20 +249,30 @@ class ImageLoader(val context: Context, fileCache: CacheService, val imageCache:
...
 }
 
 object ImageLoader {
 
+  def apply(context: Context,
+            fileCache: CacheService,
+            imageCache: MemoryImageCache,
+            bitmapLoader: BitmapDecoder,
+            permissions: PermissionsService,
+            assetLoader: AssetLoader) = {
```
deanrobertcook: The apply methods on Services and other (semi) singleton entities are not really necessary. In fact, I would generally prefer having `new XYZService()` to distinguish these services from the data which we pass around inside of them, which are usually case classes anyway. It also adds an extra object and a handful of methods without any real gain.

zmessaging/src/main/scala/com/waz/service/images/ImageLoader.scala
```diff
@@ -36,39 +36,37 @@ import com.waz.threading.{CancellableFuture, Threading}
...
 import com.waz.utils.IoUtils._
-import com.waz.utils.wrappers.URI
+import com.waz.utils.wrappers.{Context, Intent, URI}
 import com.waz.utils.{IoUtils, Serialized, returning}
 import com.waz.{PermissionsService, bitmap}
 
 import scala.concurrent.Future
 
-class ImageLoader(val context: Context, fileCache: CacheService, val imageCache: MemoryImageCache,
+class ImageLoader(val context: Context, fileCache: CacheService, val memoryCache: MemoryImageCache,
```
deanrobertcook: You should be able to get rid of the `val` in front of `memoryCache`, and I just checked, you can probably also get rid of the `val` in front of `context` too.

zmessaging/src/main/scala/com/waz/ui/MemoryImageCache.scala
```diff
@@ -24,28 +24,23 @@ import com.waz.bitmap
 import com.waz.model.AssetId
 import com.waz.threading.{CancellableFuture, Threading}
 import com.waz.ui.MemoryImageCache.BitmapRequest.{Regular, Round, Single}
-import com.waz.utils.TrimmingLruCache
+import com.waz.utils.{Cache, TrimmingLruCache}
 import com.waz.utils.TrimmingLruCache.CacheSize
+import com.waz.utils.wrappers.Bmp
 
-class MemoryImageCache(val context: Context) {
+class MemoryImageCache(val context: Context, private val lru: Cache[MemoryImageCache.Key, MemoryImageCache.Entry]) {
```
deanrobertcook: I think `private val` is somewhat redundant? You can just get rid of modifier altogether. 

Also, I don't see that `context` is used anywhere.

macrosupport/src/main/scala/com/waz/ZLog.scala
```diff
@@ -58,7 +58,7 @@ private object ZLogMacros {
     import c.universe._
     q"""
         if (com.waz.utils.isTest) {
-          if (com.waz.ZLog.testLogLevel <= com.waz.ZLog.LogLevel.Error) ${reify(println(s"E: ${tag.splice}: ${message.splice}"))}
+          if (com.waz.ZLog.testLogLevel <= com.waz.ZLog.LogLevel.Error) ${reify(println(s"E: ${tag.splice}: ${message.splice}: ${cause.splice}"))}
```
deanrobertcook: Probably don't need that - it didn't really work.

tests/unit/src/test/scala/com/waz/bitmap/BitmapDecoderSpec.scala
```diff
@@ -18,17 +18,17 @@
...
 
 import android.graphics.Bitmap
-import org.robolectric.Robolectric
 import org.scalatest.{FeatureSpec, Matchers, RobolectricTests}
+import com.waz.utils.wrappers.Bmp
 
 class BitmapDecoderSpec extends FeatureSpec with Matchers with RobolectricTests {
 
-  lazy val bitmap = Bitmap.createBitmap(10, 10, Bitmap.Config.ARGB_8888)
+  lazy val bitmap: Bmp = Bitmap.createBitmap(10, 10, Bitmap.Config.ARGB_8888)
```
deanrobertcook: is the type necessary?

tests/unit/src/test/scala/com/waz/model/AssetDataSpec.scala
```diff
@@ -22,7 +22,7 @@ import com.waz.model.AssetMetaData.Loudness
 import com.waz.model.AssetStatus.{UploadInProgress, UploadNotStarted}
 import com.waz.specs.AndroidFreeSpec
 import com.waz.testutils.Matchers._
-import com.waz.utils.sha2
+//import com.waz.utils.sha2
```
deanrobertcook: remove

tests/unit/src/test/scala/com/waz/model/AssetDataSpec.scala
```diff
@@ -83,24 +98,17 @@ class AssetDataSpec
...
-    scheme <- Gen.oneOf("file", "content", "http")
-    path <- alphaNumStr
-  } yield URI.parse(s"$scheme://$path"))
-  def sideEffect[A](f: => A): Gen[A] = Gen.resultOf[Unit, A](_ => f)
   implicit lazy val arbConvId: Arbitrary[ConvId]         = Arbitrary(sideEffect(ConvId()))
-  implicit lazy val arbRConvId: Arbitrary[RConvId]       = Arbitrary(sideEffect(RConvId()))
   implicit lazy val arbUserId: Arbitrary[UserId]         = Arbitrary(sideEffect(UserId()))
-  implicit lazy val arbRAssetDataId: Arbitrary[RAssetId] = Arbitrary(sideEffect(RAssetId()))
-  implicit lazy val arbAssetId: Arbitrary[AssetId]       = Arbitrary(sideEffect(AssetId()))
+  implicit lazy val arbRAssetDataId: Arbitrary[RAssetId] = Arbitrary(sideEffect(RAssetId()))*/
```
deanrobertcook: Can we remove these commented out generators?

tests/unit/src/test/scala/com/waz/service/assets/DownloaderServiceSpec.scala
```diff
@@ -193,6 +196,105 @@ class DownloaderServiceSpec extends FeatureSpec with Matchers with BeforeAndAfte
...
+  val database = stub[Database]
+  val cacheStorage = stub[CacheStorage]
+  val cacheService = new CacheService(context, database, cacheStorage)
+  val result = new CacheEntry(CacheEntryData(CacheKey("...")), cacheService)
+
+  feature("Download retries") {
+    scenario("download with no retries") {
+      val fakeD = fakeDownload()
+
+      val downloader = new service.downloads.DownloaderService(testContext, global.cache, global.prefs, global.network){
```
deanrobertcook: You can pull the initialization of the Downloader service out into a method.

zmessaging/src/main/scala/com/waz/bitmap/BitmapDecoder.scala
```diff
@@ -55,29 +52,29 @@ class BitmapDecoder(context: Context) {
...
   }
-
-  def apply(resourceId: Int, inSampleSize: Int, orientation: Int): CancellableFuture[Bitmap] = dispatcher {
+/*
+  def apply(resourceId: Int, inSampleSize: Int, orientation: Int): CancellableFuture[Bmp] = dispatcher {
     retryOnError(inSampleSize, inSampleSize * 3) { sampleSize =>
       withFixedOrientation(BitmapFactory.decodeResource(context.getResources, resourceId, factoryOptions(sampleSize)), orientation)
     }
-  }
+  }*/
```
deanrobertcook: If we don't need it, drop it :)

zmessaging/src/main/scala/com/waz/service/images/ImageLoader.scala
```diff
@@ -125,49 +124,38 @@ class ImageLoader(val context: Context, fileCache: CacheService, val imageCache:
...
+        warn("permission to save image to gallery denied")
+        Future successful None
+      },
+      Future {
+        val newFile = AssetService.saveImageFile(mime)
+        IoUtils.copy(data.inputStream, new FileOutputStream(newFile))
+        val uri = URI.fromFile(newFile)
+        context.sendBroadcast(Intent.scanFileIntent(uri))
+        Some(uri)
+      }(Threading.IO)
```
deanrobertcook: Formatting is a bit off here

zmessaging/src/main/scala/com/waz/utils/wrappers/Bmp.scala
```diff
@@ -0,0 +1,56 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.utils.wrappers
+
+import android.graphics.Bitmap
+import com.waz.ui.Images
+
+import scala.language.implicitConversions
+
+trait Bmp {
```
deanrobertcook: I think the name `Bmp` might be a little hard to remember in the future. Might be better to use `Bitmap` and wherever we use the android version, import it with an alias.

macrosupport/src/main/scala/com/waz/ZLog.scala
makingthematrix: Removed.

tests/unit/src/test/scala/com/waz/bitmap/BitmapDecoderSpec.scala
makingthematrix: We need to state somehow that we want this to be a wrapper. It can be done either this way, so then the implicit conversion will be made, or we can explicitly state `lazy val bitmap = Bmp(Bitmap.createBitmap(...))`

tests/unit/src/test/scala/com/waz/model/AssetDataSpec.scala
makingthematrix: Done

tests/unit/src/test/scala/com/waz/model/AssetDataSpec.scala
makingthematrix: Yes. Done.

tests/unit/src/test/scala/com/waz/service/assets/DownloaderServiceSpec.scala
makingthematrix: Done.

zmessaging/src/main/scala/com/waz/bitmap/BitmapDecoder.scala
makingthematrix: Done.

zmessaging/src/main/scala/com/waz/service/images/ImageLoader.scala
makingthematrix: Indentation removed.

zmessaging/src/main/scala/com/waz/utils/wrappers/Bmp.scala
makingthematrix: Done.

PR #114, opened 2017-04-27
Introduced new Unknown network mode

hakonbo:
No longer assume offline on startup, but set mode to Unknown. This will help UI to not briefly display the offline indicator at app startup.


hakonbo:
Sure, fair point. Changed calls to `isOffline` to `!isOnline`

PR #117, opened 2017-04-28
WIP: Firebase upgrade

deanrobertcook:
Includes changes to simplify handling of tokens in the new `PushTokenService`. Note that the fallback logic is not yet implemented there, but please check the `PushTokenServiceSpec` for tests of the code.


zmessaging/src/main/scala/com/waz/cache/CacheService.scala
```diff
@@ -158,7 +158,7 @@ class CacheService(context: Context, storage: Database, val cacheStorage: CacheS
 
   def remove(entry: CacheEntry): Future[Unit] = {
     verbose(s"remove($entry)")
-    cacheStorage.remove(entry.data)
+    cacheStorage.remove(entry.data.key)
```
wire-android: Is this change related to this PR, seems to be more general fix, maybe it should go separately.

tests/utils/src/main/scala/com/waz/specs/AndroidFreeSpec.scala
```diff
@@ -43,7 +45,13 @@ trait AndroidFreeSpec extends BeforeAndAfterAll { this: Suite =>
...
     Intent.setUtil(JVMIntentUtil)
 
-    Threading.setUi(new SerialDispatchQueue(name = "TestUIThread"))
+    Threading.setUi(new SerialDispatchQueue({
+      Threading.executionContext(Executors.newSingleThreadExecutor(new ThreadFactory {
+        override def newThread(r: Runnable) = {
+          new Thread(r, Threading.testUiThreadName)
+        }
+      }))
+    }, Threading.testUiThreadName))
```
wire-android: This commit seems to be more general test fix, it would be better to move it to separate PR. It could be merged sooner and not pollute this PR.

zmessaging/src/main/scala/com/waz/threading/Threading.scala
```diff
@@ -71,11 +73,18 @@ object Threading {
...
+    case None => returning(new UiDispatchQueue)(setUi)
   }
 
   def setUi(ui: DispatchQueue) = this._ui = Some(ui)
 
+  val testUiThreadName = "TestUiThread"
+  def isUiThread = try {
+    Thread.currentThread() == Looper.getMainLooper.getThread
+  } catch {
+    case NonFatal(e) => Thread.currentThread().getName.contains(testUiThreadName)
```
wire-android: this `catch` looks weird, are you assuming that `Looper.getMainLooper` will always throw in test code?
what if it is used in Robolectric tests?
could we do more explicit check if this is a test code?

zmessaging/src/main/AndroidManifest.xml
```diff
@@ -6,24 +6,20 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
 
-    <!-- required for GCM !-->
-    <uses-permission android:name="android.permission.WAKE_LOCK" />
```
wire-android: wake locks are also used directly in our code
we should probably leave this permission and only remove misleading comment

zmessaging/src/main/scala/com/waz/utils/wrappers/GoogleApi.scala
```diff
@@ -0,0 +1,40 @@
...
+  override val isGooglePlayServicesAvailable = Signal[Boolean](false)
+
+  override def checkGooglePlayServicesAvailable(activity: Activity) = api.isGooglePlayServicesAvailable(activity) match {
+    case ConnectionResult.SUCCESS => isGooglePlayServicesAvailable ! true
+    case code if api.isUserResolvableError(code) => api.getErrorDialog(activity, code, requestGooglePlayServices)
+    case code => warn(s"Google Play Services not available: error code: $code")
+  }
+
+  override def onActivityResult(requestCode: Int, resultCode: Int) = requestCode match {
+    case `requestGooglePlayServices` =>
```
wire-android: do we really need quotes here?
it shouldn't be needed with constants

zmessaging/src/main/scala/com/waz/utils/wrappers/GoogleApi.scala
```diff
@@ -0,0 +1,40 @@
...
+      resultCode match {
+        case Activity.RESULT_OK => isGooglePlayServicesAvailable ! true
+        case _ => warn("Failed to update/install Google Play Services")
+      }
+    case _ => //
+  }
+}
+
+object GoogleApi {
+  val requestGooglePlayServices = 7976
```
wire-android: might be better to use upper case, per scala style guidelines

deanrobertcook:
Test-only improvements moved to https://github.com/wireapp/wire-android-sync-engine/pull/124

zmessaging/src/main/scala/com/waz/threading/Threading.scala
deanrobertcook: This is only going to be used in the `AndroidFreeSpec`, where I don't start any robolectric contexts (so any calls to Android classes will fail) and we replace the default `UIDispatchQueue` with a different one that works on the JVM.

 In robolectric, I think the `getMainLooper` class should succeed, so those tests should be unaffected.

zmessaging/src/main/AndroidManifest.xml
deanrobertcook: true - good catch!

tests/utils/src/main/scala/com/waz/specs/AndroidFreeSpec.scala
deanrobertcook: this is now in https://github.com/wireapp/wire-android-sync-engine/pull/124

zmessaging/src/main/scala/com/waz/cache/CacheService.scala
deanrobertcook: Now in https://github.com/wireapp/wire-android-sync-engine/pull/124

zmessaging/src/main/scala/com/waz/model/AccountData.scala
```diff
@@ -112,20 +122,21 @@ object AccountData {
...
+    val Cookie = opt(text[Cookie]('cookie, _.str, AuthenticationManager.Cookie))(_.cookie)
     val Token = opt(text[Token]('access_token, JsonEncoder.encodeString[Token], JsonDecoder.decode[Token]))(_.accessToken)
     val UserId = opt(id[UserId]('user_id)).apply(_.userId)
     val ClientId = opt(id[ClientId]('client_id))(_.clientId)
     val ClientRegState = text[ClientRegistrationState]('reg_state, _.name(), ClientRegistrationState.valueOf)(_.clientRegState)
-    val Handle = opt(handle('handle))(_.handle)
     val PrivateMode = bool('private_mode)(_.privateMode)
 
     override val idCol = Id
     override val table = Table("Accounts", Id, Email, Hash, EmailVerified, Cookie, Phone, Token, UserId, ClientId, ClientRegState, Handle, PrivateMode)
```
deanrobertcook: Yes, migration is something I still need to do. Thanks for reminding me :)

PR #123, opened 2017-05-04
makes CachedStorage return an asset instead of None if the update was not necessary

makingthematrix:
Until recently comparing two AssetData always returned false which forced unnecessary updates in CachedStorage. I changed that but it uncovered another error: When two AssetData were the same, the update method returned None, whereas everyone expected to have a tuple: the old and the new AssetData (even if they were the same). This PR fixes that.


wire-android:
This is quite dangerous, there might be several places relying on on this `None` result.
This change may lead to infinite update loops, or excessive events being generated (events to UI or to backend). 
I'm not sure if there still are places like that, but for sure there were some time ago.
Make sure to check every place where update is called. This is very low level api and might be used everywhere, not just on assets.

PR #124, opened 2017-05-05
Test improvements and Preferences tidy up

deanrobertcook:
Necessary changes to allow better mocking and testing in https://github.com/wireapp/wire-android-sync-engine/pull/117

Includes, among other things:
* Unifying Global and User preferences under a common trait
* In-memory `Preferences` for tests 
* Migrating UI and ZMS preferences to one file
* Wrapping Localytics/HockeyApp to prevent calls to Android methods
* A single-thread ExecutionContext to act as the "UI" thread in Android-free tests
* Adding a trait for the `CachedStorage` class to allow mocking of `Storage` level classes.


zmessaging/src/main/scala/com/waz/service/push/GcmGlobalService.scala
```diff
@@ -130,16 +126,22 @@ object GcmGlobalService {
...
+      token   <- prefs.preference[String](RegistrationIdPref).apply()
+      userId  <- prefs.preference[String](RegistrationUserPref).apply()
+      version <- prefs.preference[Int](RegistrationVersionPref).apply()
+    } yield GcmRegistration(token, AccountId(userId), version)
+
+    def update(reg: GcmRegistration)(implicit ec: ExecutionContext, prefs: Preferences) = for {
+      _ <- prefs.preference[String](RegistrationUserPref) := reg.token
+      _ <- prefs.preference[String](RegistrationUserPref) := reg.user.str
+      _ <- prefs.preference[Int](RegistrationVersionPref) := reg.version
+    } yield {}
```
wire-android: would be nice to replace that with single preference using PrefCodec[GcmRegistration], or as you mentioned earlier use json

zmessaging/src/main/scala/com/waz/service/push/GcmService.scala
```diff
@@ -130,20 +131,20 @@ class GcmService(accountId:         AccountId,
 
   shouldReRegister {
     case true =>
-      (for {
+      for {
         retries <- registrationRetryCount()
-        if retries > retryFailLimit
```
wire-android: are you sure we don't need to use `retryFailLimit`?
will this retry forever?

zmessaging/src/main/scala/com/waz/utils/CachedStorageImpl.scala
```diff
@@ -51,7 +51,57 @@ object StorageDao {
   }
 }
 
-class CachedStorage[K, V](cache: LruCache[K, Option[V]], db: Database)(implicit val dao: StorageDao[K, V], tag: LogTag = "CachedStorage") {
+trait CachedStorage[K, V] {
+
+  def onAdded: EventStream[Seq[V]]
+  def onUpdated: EventStream[Seq[(V, V)]]
+  def onDeleted: EventStream[Seq[K]]
```
wire-android: In implementation those need to be `vals`, maybe we should enforce that by declaring those as `val` in this trait?

zmessaging/src/main/scala/com/waz/service/push/GcmGlobalService.scala
deanrobertcook: This will be deleted soon by the new PushTokenService, this was just so the PR compiled

zmessaging/src/main/scala/com/waz/utils/CachedStorageImpl.scala
deanrobertcook: yup, true

zmessaging/src/main/scala/com/waz/content/Preferences.scala
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }
+}
+
+/**
+  * Global preference based on Android SharedPreferences. Note, here we need to save preferences to the correct primitive
+  * type in SharedPreferences, as the Android system uses these types by default (e.g., a CheckBoxPreference defined in XML
+  * will store a boolean preference in the shared prefs document
+  */
+class GlobalPreferences(context: Context) extends Preferences {
```
makingthematrix: This is the Android `context` and a few lines down it is used during the instance construction to set `prefs`. In order to make it easier to use in the future in Android-free environment I'd suggest to turn it into:
```
class GlobalPreferences(context: Context, protected val prefs: SharedPreferences) {
...
}
...
object GlobalPreferences {
  def apply(context: Context) = {
    val globalPrefs = new GlobalPreferences(context, context.getSharedPreferences( ... ))
    globalPrefs.migrate() // also please move 'migrate' here
    globalPrefs
  }
}
```
It's easier to read too: You don't have to look through the code of the class to check if something happens during the initialization. Everything's in the 'apply' method.

EDIT: Or maybe even `class GlobalPreferences(context:Context, protected val prefs: Option[SharedPreferences] = None` if you use need GlobalPreferences in a test but do not expect it to be used.

PR #125, opened 2017-05-05
AN-5163 bugfix. Android-free unit tests in AssetSyncHandler. 

makingthematrix:
The current version of AN-5163 fix prevents the receiver from displaying a dark container and the sender from displaying an invalid "Delivered" footer. However, the sender has to tap "Retry" in order to retry sending. It would be nice to have automatic retries.

### DIFF
--- a/tests/unit/src/test/scala/com/waz/sync/handler/AssetSyncHandlerSpec.scala
+++ b/tests/unit/src/test/scala/com/waz/sync/handler/AssetSyncHandlerSpec.scala
@@ -1,0 +1,118 @@
+/*
+ * Wire
+ * Copyright (C) 2016 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.sync.handler
+
+import java.io.{File, InputStream}
+
+import com.waz.cache._
+import com.waz.model.AssetData.RemoteData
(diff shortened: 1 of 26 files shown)


tests/unit/src/test/scala/com/waz/sync/handler/AssetSyncHandlerSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+import org.scalamock.scalatest.MockFactory
+import org.scalatest.{BeforeAndAfter, FeatureSpec, Matchers}
+
+import scala.annotation.tailrec
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.duration._
+
+/**
+  * Created by admin on 03/05/17.
+  */
```
deanrobertcook: Remove comments - You can remove this in the template for new files in IntelliJ

zmessaging/src/main/scala/com/waz/cache/CacheService.scala
```diff
@@ -27,18 +27,72 @@ import com.waz.cache.CacheEntryData.CacheEntryDao
...
 
-class CacheService(context: Context, storage: Database, val cacheStorage: CacheStorage) {
+trait CacheService {
+  def createManagedFile(key: Option[AESKey] = None)(implicit timeout: Expiration = CacheService.DefaultExpiryTime): CacheEntry
+  def createForFile(key: CacheKey = CacheKey(),
+                    mime: Mime = Mime.Unknown,
+                    name: Option[String] = None,
+                    cacheLocation: Option[File] = None,
+                    length: Option[Long] = None)
+                   (implicit timeout: Expiration = CacheService.DefaultExpiryTime): Future[CacheEntry]
```
deanrobertcook: Maybe line the types and defaults up too, then it's really easy to read :)

zmessaging/src/main/scala/com/waz/cache/CacheService.scala
```diff
@@ -27,18 +27,72 @@ import com.waz.cache.CacheEntryData.CacheEntryDao
...
+
+  def deleteExpired(): CancellableFuture[Unit]
+  def optSignal(cacheKey: CacheKey): Signal[Option[CacheEntry]]
+}
+
+class CacheServiceImpl(context: Context, storage: Database, cacheStorage: CacheStorage) extends CacheService {
   import CacheService._
   import Threading.Implicits.Background
 
+  private implicit val logTag: LogTag = logTagFor[CacheService]
```
deanrobertcook: `import com.waz.ZLog.ImplicitTag._` instead - these are old and generally I'm trying to remove them

PR #127, opened 2017-05-10
Firebase Upgrade

deanrobertcook:


zmessaging/src/main/scala/com/waz/sync/client/PushTokenClient.scala
```diff
@@ -43,33 +42,34 @@ class GcmClient(netClient: ZNetClient) {
     netClient.updateWithErrorHandling("deletePushToken", Request.Delete[Unit](s"$PushesPath/$token"))
 }
 
-object GcmClient {
+object PushTokenClient {
   val PushesPath = "/push/tokens"
 
-  case class GcmToken(token: String, app: GcmSenderId, clientId: ClientId, transport: String = "GCM")
-  object GcmToken {
+  //TODO shoiuld transport be FCM?
```
wire-android: what about this todo?

zmessaging/src/main/scala/com/waz/zms/FCMHandlerService.scala
```diff
@@ -0,0 +1,183 @@
...
+        case _ => warn("No zms instance available"); Future.successful({})
+      }
+    }
+  }
+
+  /**
+    * Called when the device hasn't connected to the FCM server in over 1 month, or there are more than 100 FCM
+    * messages available for this device on the FCM servers.
+    *
+    * TODO does a syncHistory make sense?
```
wire-android: Don't think `syncHistory` gives us anything here. It would be called on push connection anyway.
Is there anything we should do on this callback after all? We have our own missing notification tracking on websocket, this callback could just be ignored.

tests/integration/src/test/scala/com/waz/api/ZMessagingApiSpec.scala
```diff
@@ -61,7 +61,7 @@ class ZMessagingApiSpec extends FeatureSpec with OptionValues with Matchers with
 
       // make sure current user data is saved in db (just to check that Robolectric is behaving correctly)
       Await.result(globalModule.accountsStorage.find(EmailCredentials(EmailAddress(email), Some(password))), timeout) match {
-        case Some(AccountData(id, e, hash, _, _, _, _, _, _, _, _, _, _)) =>
+        case Some(AccountData(id, e, hash, _, _, _, _, _, _, _, _, _, _, _)) =>
```
makingthematrix: Just a tip: In such situation you might want to switch to `case Some(data)` and then use `data.id`, `data.email` and `data.hash`. The test obviously does not care about other fields of the account data, nor their order, so fixing it every time something changes in `AccountData` is just unnecessary work.

zmessaging/src/main/scala/com/waz/service/otr/OtrServiceImpl.scala
```diff
@@ -288,8 +292,8 @@ class OtrService(selfUserId: UserId, clientId: ClientId, val clients: OtrClients
 
 }
 
-object OtrService {
-  private implicit val logTag: LogTag = logTagFor[OtrService]
+object OtrServiceImpl {
```
makingthematrix: `object OtrService` may still have this name. `OtrServiceImpl` is the default implementation of `OtrService`, but these few methods from the object seem to be common to all classes extending the `OtrService` trait.

zmessaging/src/main/scala/com/waz/service/push/PushServiceImpl.scala
```diff
@@ -173,7 +182,7 @@ class PushService(context: Context, keyValue: UserPreferences, client: EventsCli
     }
 }
 
-object PushService {
+object PushServiceImpl {
```
makingthematrix: Same thing as with `OtrService`: The name of the object could stay the same.

zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
```diff
@@ -51,7 +51,7 @@ import scala.concurrent.Future
 import scala.concurrent.Future.successful
 
 class MessagesSyncHandler(context: Context, service: DefaultMessagesService, msgContent: MessagesContentUpdater, convEvents: ConversationEventsService,
-                          client: MessagesClient, otr: OtrService, otrSync: OtrSyncHandler, convs: DefaultConversationsContentUpdater, storage: MessagesStorage,
+                          client: MessagesClient, otr: OtrServiceImpl, otrSync: OtrSyncHandler, convs: DefaultConversationsContentUpdater, storage: MessagesStorage,
```
makingthematrix: I think it should be `OtrService` here.

zmessaging/src/main/scala/com/waz/sync/handler/OpenGraphSyncHandler.scala
```diff
@@ -39,7 +39,7 @@ import org.threeten.bp.Instant
 
 import scala.concurrent.Future
 
-class OpenGraphSyncHandler(convs: ConversationStorage, messages: MessagesStorage, otrService: OtrService, assetSync: AssetSyncHandler, assetsStorage: AssetsStorage, otrSync: OtrSyncHandler, client: OpenGraphClient, imageGenerator: ImageAssetGenerator, imageLoader: ImageLoader, assetClient: AssetClient) {
+class OpenGraphSyncHandler(convs: ConversationStorage, messages: MessagesStorage, otrService: OtrServiceImpl, assetSync: AssetSyncHandler, assetsStorage: AssetsStorage, otrSync: OtrSyncHandler, client: OpenGraphClient, imageGenerator: ImageAssetGenerator, imageLoader: ImageLoader, assetClient: AssetClient) {
```
makingthematrix: And here.

zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
deanrobertcook: I guess in these spots we would need to ensure all necessary methods are in the trait. Until we get around to testing the `MessagesSyncHandler`, and we add those methods, then it's fine to leave the dependency on the Impl

tests/unit/src/test/scala/com/waz/db/ZGlobalDBSpec.scala
```diff
@@ -98,8 +98,8 @@ class ZGlobalDBSpec extends FeatureSpec with Matchers with OptionValues with Ins
...
+          AccountData(AccountId("8546c628-c9e8-45d6-82dd-7f6dcb56e171"), Some(EmailAddress("joachim.hofer+001@wearezeta.com")), "Xx/rjrJc0B/MhvaAt/aegKrs+bohYNkBTnZ3wJbl+Pg=", None, handle = Some(Handle()), None, verified = true, Some(Cookie("nK4NNJ7XN9-riGCcJ6YDCIXYEpHSYJWV2L9s3at1brf33Nb5TcFjY341iQHhQ7GjAS8sDgfXNx6NvzmSyXDXBQ==.v=1.k=1.d=1458844442.t=u.l=.u=e222adf6-22a0-4180-b628-936049f0899f.r=ccaae5e6")), userId = Some(userId1)),
+          AccountData(AccountId("09621ddd-736f-4ec5-b4b5-d24cbb56b9f3"), Some(EmailAddress("joachim.hofer+003@wearezeta.com")), "WtjSXe7G8CHlcy4PRxGoaisUr9UGyKR51zriDwIFAco=", Some(PhoneNumber("+0123456789")), handle = Some(Handle()), None, verified = true, Some(Cookie("u0mEC2etISwrAAf-_pNwG204HG5-Uf7EIRFFTp1TEqKGcSIXDbFC9_i8PftnKRTWSjUsAbZ-PHVIxS3eZDK-AQ==.v=1.k=1.d=1459026632.t=u.l=.u=9a01b792-42f6-4dee-a3c0-e22179d742f8.r=591684f6")), userId = Some(userId2))
         )
       }
```
deanrobertcook: @wire-android I started looking into writing a test for this, but it seems the Migration tests are broken (they don't even open an SQLite database). I started trying to copy what I've done in the scala utils repo so I can run tests on the JVM with a different implementation of SQLite, but it's going to be a very big task.

Using stetho, I've checked and confirmed everything gets saved properly.

zmessaging/src/main/java/com/waz/model/sync/SyncCommand.java
```diff
@@ -43,10 +43,10 @@
     PostConvJoin("post-conv-join"),
     PostConvLeave("post-conv-leave"),
     PostConnection("post-connection"),
-    DeleteGcmToken("delete-gcm-token"),
+    DeletePushToken("delete-push-token"),
     SyncRichMedia("sync-rich-media"),
     SyncConnectedUsers("sync-connected"),
-    RegisterGcmToken("register-gcm-token"),
+    RegisterPushToken("register-push-token"),
```
deanrobertcook: Actually, this was the case I really wanted migration tests for... will still try and fix them before merging this PR...

tests/unit/src/test/scala/com/waz/service/push/WebSocketClientServiceSpec.scala
```diff
@@ -45,70 +45,69 @@ class WebSocketClientServiceSpec extends FeatureSpec with Matchers with Robolect
...
+//      awaitUi(50.millis)
+//
+//      error shouldEqual false
+//
+//      withDelay {
+//        service.connectionError.currentValue shouldEqual Some(true)
+//        error shouldEqual true
+//      }
+//    }
+//  }
```
deanrobertcook: Will try and fix. I guess I can add a TODO

PR #136, opened 2017-05-18
AsyncClient refactoring and android-free user tests.

makingthematrix:
There are now wrappers on Koushi's AsyncHttpClient, AsyncHttpRequest and AsyncHttpResponse, and worker classes taking care of processing the request and the response, so they can be mocked in tests. The Request and Response case classes can be converted to respective AsyncHttp* classes and so they are used directly in AsyncClient.


makingthematrix:
build again

zmessaging/src/main/scala/com/waz/znet/ClientWrapper.scala
```diff
@@ -21,101 +21,146 @@ import java.security.cert.X509Certificate
...
+    new Thread(new Runnable {
+      override def run(): Unit = p.complete(Try{ f.get })
+    }).start
+
+    new CancellableFuture(p){
+      override def cancel()(implicit tag: LogTag): Boolean = f.cancel(true)
+    }
+  }
+
+  def toKoushi(wrapper: ClientWrapper): AsyncHttpClient = wrapper match {
```
wire-android: Not fan of this naming. Would rather prefer to use project name, not the author's. 
And even if you think his name is better then we should spell it correctly.

zmessaging/src/main/scala/com/waz/znet/ClientWrapper.scala
```diff
@@ -21,101 +21,146 @@ import java.security.cert.X509Certificate
...
-object ClientWrapper extends ClientWrapper {
+object ClientWrapper {
   import Threading.Implicits.Background
 
+  import scala.language.implicitConversions
+
+  implicit def koushiToCancellable[T](f: KFuture[T]): CancellableFuture[T] = {
+    val p = Promise[T]()
+
+    new Thread(new Runnable {
```
wire-android: Please don't create new threads, if you really need blocking operation than use IO background queue (or some special executor).
 
AndroidAsync futures also support callbacks (`KFuture.setCallback`), so there is no need to use blocking `get` here.

tests/integration/src/test/scala/com/waz/znet/AsyncClientWebSpec.scala
```diff
@@ -109,7 +111,7 @@ class AsyncClientWebSpec extends FeatureSpecLike with Matchers with BeforeAndAft
     scenario("post data") {
       val data = (0 to 100).map("test_" + _).mkString(", ").getBytes("utf8")
 
-      Await.result(client(URI.parse("http://posttestserver.com/post.php"), "POST", new BinaryRequestContent(data, "text/plain"), timeout = AsyncClient.DefaultTimeout), 45.seconds) match {
+      Await.result(client(URI.parse("http://posttestserver.com/post.php"), Request.Post("/post.php", new BinaryRequestContent(data, "text/plain"))), 45.seconds) match {
```
makingthematrix: OK. I made changes to `Request`. Now the uris are stored inside it as baseUri and resourcePath.

zmessaging/src/main/scala/com/waz/znet/ClientWrapper.scala
makingthematrix: I think we should talk about it.

zmessaging/src/main/scala/com/waz/znet/ClientWrapper.scala
makingthematrix: OK, changed.

PR #139, opened 2017-05-24
[AN-5153] Receiving AssetV3 image does not restore deleted conversation

makingthematrix:
Receiving a location neither. Fixed.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/conversation/UnarchivingEventProcessingStage.scala
+++ b/zmessaging/src/main/scala/com/waz/service/conversation/UnarchivingEventProcessingStage.scala
@@ -17,8 +17,6 @@
  */
 package com.waz.service.conversation
 
-import com.waz.ZLog._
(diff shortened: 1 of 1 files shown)


zmessaging/src/main/scala/com/waz/service/conversation/UnarchivingEventProcessingStage.scala
```diff
@@ -58,11 +56,13 @@ object UnarchivingEventProcessingStage {
     case MemberLeaveEvent(_, _, _, leaving) if leaving contains selfUserId => false
     case GenericMessageEvent(_, _, _, GenericMessage(_, content)) =>
       content match {
-        case _: Text  => true
-        case _: ImageAsset => true
-        case _: Knock => true
-        case _: Ephemeral => true
-        case _        => false
+        case _: Text        => true
+        case _: ImageAsset  => true
```
makingthematrix: Do we use it still?

PR #141, opened 2017-05-27
Create Apache HTTPClient for running integration tests on JVM

deanrobertcook:


tests/unit/src/test/scala/com/waz/znet/AsyncClientSpec.scala
```diff
@@ -189,7 +189,7 @@ class AsyncClientSpec extends AndroidFreeSpec {
     Await.result(client(request), waitTime)
   }
 
-  private def doPost[A: ContentEncoder](path: String, data: A, auth: Boolean = false, timeout: FiniteDuration = AsyncClient.DefaultTimeout, waitTime: FiniteDuration = 500.millis, client: AsyncClient = client) = {
+  private def doPost[A: ContentEncoder](path: String, data: A, auth: Boolean = false, timeout: FiniteDuration = AsyncClient.DefaultTimeout, waitTime: FiniteDuration = 500.millis, client: AsyncClientImpl = client) = {
```
makingthematrix: `AsyncClient` should be enough, as it declares `apply`.

tests/utils/src/main/scala/com/waz/testutils/ApacheHttpClient.scala
```diff
@@ -0,0 +1,78 @@
...
+  */
+class ApacheHttpClient extends AsyncClient {
+
+  import com.waz.threading.Threading.Implicits.Background
+
+  override def apply(request: Request[_]) = {
+    debug(s"Starting request[${request.httpMethod}](${request.absoluteUri}) with body: ${request.getBody}, headers: '${request.headers}'")
+    val juri = request.absoluteUri.getOrElse(throw new IllegalArgumentException("No uri specified")).asInstanceOf[JavaURI].uri
+    CancellableFuture {
+      val resp = buildHttpRequest(juri, request.httpMethod, request.getBody, request.headers, request.followRedirect, request.timeout).execute().returnResponse()
```
makingthematrix: I think this `resp` should be called `req`.

zmessaging/src/main/scala/com/waz/service/GlobalModule.scala
```diff
@@ -73,7 +73,7 @@ class GlobalModule(val context: Context, val backend: BackendConfig) { global =>
   lazy val tempFiles: TempFileService                            = wire[TempFileService]
 
   lazy val clientWrapper: Future[ClientWrapper] = ClientWrapper()
-  lazy val client: AsyncClient = new AsyncClient(decoder, AsyncClient.userAgent(metadata.appVersion.toString, ZmsVersion.ZMS_VERSION), clientWrapper)
+  lazy val client: AsyncClientImpl = new AsyncClientImpl(decoder, AsyncClient.userAgent(metadata.appVersion.toString, ZmsVersion.ZMS_VERSION), clientWrapper)
```
makingthematrix: Here too `AsyncClient` should be enough (or no explicit type at all)

zmessaging/src/main/scala/com/waz/znet/ZNetClient.scala
```diff
@@ -47,7 +47,7 @@ class ZNetClient(credentials: CredentialsHandler,
                  loginClient: LoginClient) {
 
   def this(global: GlobalModule, email: String, passwd: String) = this(new BasicCredentials(EmailAddress(email), Some(passwd)), global.client, global.backend, global.loginClient)
-  def this(email: String, passwd: String, client: AsyncClient) =
+  def this(email: String, passwd: String, client: AsyncClientImpl) =
```
makingthematrix: `AsyncClient` should be enough

tests/utils/src/main/scala/com/waz/testutils/ApacheHttpClient.scala
deanrobertcook: The return type is an Apache Response, it performs the request immediately after building the request

tests/mocked/src/test/scala/com/waz/mocked/MockedClientSuite.scala
```diff
@@ -121,7 +121,7 @@ trait MockedClientSuite extends ApiSpec with MockedClient with MockedWebSocket w
 
     override lazy val websocket: service.push.WebSocketClientService = new service.push.WebSocketClientService(context, lifecycle, zNetClient, network, global.backend, clientId, timeouts, pushToken) {
 
-      override def createWebSocketClient(clientId: ClientId): WebSocketClient = new WebSocketClient(context, zNetClient.client, Uri.parse(backend.websocketUrl), zNetClient.auth) {
+      override def createWebSocketClient(clientId: ClientId): WebSocketClient = new WebSocketClient(context, zNetClient.client.asInstanceOf[AsyncClientImpl], Uri.parse(backend.websocketUrl), zNetClient.auth) {
```
deanrobertcook: I don't know how I feel about over-complicating the `AsyncClient` trait - I think the problem is that the `WebSocketClient` knows too much about the internals of the `AyncClientImpl`, namely the `userAgent` and `ClientWrapper`, so I left it like this to reflect that. 

I'd be happy with having a `webSocket` method on the `AsyncClient` trait, but still, it's only used in the `WebsocketClient` and is still too dependent on the `AndroidAsync` library.

PR #142, opened 2017-05-29
Fix ssl handshake on assets

wire-android:
Added a hack to enable SNI extension in SSLEngine, this is required by our backend (on assets endpoint).
Updated to android api 24 as it has methods to set SNI directly.

Tested this fix only on android M and N, so it needs a bit more testing before release.


deanrobertcook:
Will create hotfix release - not merging

PR #143, opened 2017-05-30
[AN-5266] deleting ephemeral videos on the sender side

makingthematrix:


zmessaging/src/main/scala/com/waz/service/assets/AssetService.scala
```diff
@@ -191,7 +192,13 @@ class AssetServiceImpl(storage: AssetsStorage, generator: ImageAssetGenerator, c
...
   def mergeOrCreateAsset(assetData: AssetData): Future[Option[AssetData]] = storage.mergeOrCreateAsset(assetData)
 
-  def removeAssets(ids: Iterable[AssetId]): Future[Unit] = storage.remove(ids)
+  def removeAssets(ids: Iterable[AssetId]): Future[Unit] = Future.traverse(ids)(removeSource).flatMap(_ => storage.remove(ids))
+
+  def removeSource(id: AssetId): Future[Unit] = for {
+    asset <- storage.get(id)
+  } yield asset.map(_.source).collect {
+    case Some(uri) if uri.getPath.startsWith("file://") => new File(uri.getPath).delete()
+  }
```
deanrobertcook: I think a simple `storage.get(id).map(_.source).foreach { ... }` might look a little nicer? To me, yields/collects kind of imply that we want to return something at first glance.

zmessaging/src/main/scala/com/waz/service/messages/EphemeralMessagesService.scala
```diff
@@ -103,11 +105,18 @@ class EphemeralMessagesService(selfUserId: UserId, messages: MessagesContentUpda
...
+        msg.copy(expired = true)
+      case ANY_ASSET =>
+        removeSource(msg)
+        msg.copy(expired = true)
+      case VIDEO_ASSET =>
+        removeSource(msg)
+        msg.copy(expired = true)
+      case AUDIO_ASSET =>
+        removeSource(msg)
+        msg.copy(expired = true)
```
deanrobertcook: `case ASSET | ANY_ASSET | VIDEO_ASSET | AUDIO_ASSET => ...`

PR #148, opened 2017-06-02
Dean-proof migrations

deanrobertcook:
This is mostly just to protect internal users from history loss when doing migrations that might result in an error. It gives us a chance to fix the error, make another release and then apply the fixed migration on the backed up database. This should never get into the wild:

Next step would be to show a warning in the UI so that Internal users report to us directly.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/db/Migration.scala
+++ b/zmessaging/src/main/scala/com/waz/db/Migration.scala
@@ -17,9 +17,12 @@
  */
 package com.waz.db
 
+import java.io.{File, FileInputStream, FileOutputStream}
+
 import android.database.sqlite.SQLiteDatabase
 import com.waz.HockeyApp
 import com.waz.ZLog._
+import com.waz.api.ZmsVersion
 import com.waz.utils.wrappers.DB
 
 import scala.util.control.NonFatal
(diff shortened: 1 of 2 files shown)


zmessaging/src/main/scala/com/waz/db/Migration.scala
```diff
@@ -89,6 +94,13 @@ class Migrations(migrations: Migration*) {
   @throws[IllegalStateException]("If no migration plan can be found for given versions")
   def migrate(storage: DaoDB, fromVersion: Int, toVersion: Int)(implicit db: SQLiteDatabase): Unit = {
     if (fromVersion != toVersion) {
+
+      //If there is an existing backup, replace the current db with it so that we apply migrations to the backed up data.
+      if (ZmsVersion.DEBUG) {
+        if (storage.backupFile.exists()) copy(storage.backupFile, new File(db.getPath))
+        else copy(new File(db.getPath), storage.backupFile)
```
wire-android: does it actually work? is it allowed?
We are overriding currently open db file, I would expect this to lead to weird errors. 
We should probably close current storage, replace the file and open it again.
Restored backup will most likely be in different version, so after the file is replaced 'fromVersion' is no longer valid.

PR #152, opened 2017-06-07
[AN-5266] ephemeral images are being deleted from the storage

makingthematrix:
Fix to AN-5266 was to delete ephemeral videos and audios form the temporary folder, but it also deleted images. If the user sent an ephemeral photo, she was not able to see it anymore in the Wire's gallery. Fixed that.


zmessaging/src/main/scala/com/waz/service/messages/EphemeralMessagesService.scala
```diff
@@ -39,7 +40,7 @@ import scala.util.Random
 
 // TODO: obfuscate sent messages when they expire
 class EphemeralMessagesService(selfUserId: UserId, messages: MessagesContentUpdater, storage: MessagesStorageImpl,
-                               db: ZmsDatabase, sync: SyncServiceHandle, assets: AssetService) {
+                               db: ZmsDatabase, sync: SyncServiceHandle, assets: AssetService, users: UserService) {
```
deanrobertcook: you already have `selfUserId` at the start of the imports there - not sure what the point of the future is.

makingthematrix: Sorry, didn't notice that.

PR #156, opened 2017-06-09
[AN-5333] Old calls briefly appear when coming back online

hakonbo:


zmessaging/src/main/scala/com/waz/service/call/AvsV3.scala
```diff
@@ -178,7 +178,7 @@ object AvsV3 {
   def instant(uint32_t: Uint32_t) = Instant.ofEpochMilli(uint32_t.value.toLong * 1000)
 
   def uint32_tTime(instant: Instant) =
-    returning(Uint32_t((Instant.now.toEpochMilli / 1000).toInt))(t => verbose(s"uint32_tTime for $instant = ${t.value}"))
+    returning(Uint32_t((instant.toEpochMilli / 1000).toInt))(t => verbose(s"uint32_tTime for $instant = ${t.value}"))
```
deanrobertcook: giphy

PR #160, opened 2017-06-15
[AN-5331] Internal log

makingthematrix:
* An InternalLog object lying just under ZLog and taking most of its work: comparing log levels, and redirecting to Android and to our own log file.
* Logging to file is buffered. The contents are flushed when they grow over 1MB or on specific events (defined in UI)
* ZLog macros work better now. If DebugMode is false they are expanded to empty strings. Before they were expanded to "if (DEBUG) ... else ..."
* A new reporter in ReportingService
* logTag assignments replaced with implicitLogTag in many places

### DIFF
--- a/macrosupport/src/main/scala/com/waz/ZLog.scala
+++ b/macrosupport/src/main/scala/com/waz/ZLog.scala
@@ -19,23 +19,13 @@ package com.waz
 
 import scala.language.experimental.macros
 import scala.reflect.macros.blackbox.Context
-import android.util.Log
 
 import scala.annotation.tailrec
 
 object ZLog {
   type LogTag = String
 
-  type LogLevel = LogLevel.Value
-  object LogLevel extends Enumeration {
-    val Verbose, Debug, Warn, Info, Error = Value
-  }
-
-  var testLogLevel: LogLevel = LogLevel.Error
-  def setTestLogging() = this.testLogLevel = LogLevel.Verbose
-  def testLogging: Boolean = this.testLogLevel == LogLevel.Verbose
-
-  @volatile var minimumLogLevel: Int = Log.VERBOSE
+  @volatile var minimumLogLevel: Int = android.util.Log.VERBOSE
 
   def logTagFor[A <: Singleton](a: A): String = macro ZLogMacros.logTagForSingleton[A]
   def logTagFor[A]: String = macro ZLogMacros.logTagFor[A]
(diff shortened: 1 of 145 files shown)


tests/unit/src/test/scala/com/waz/log/InternalLogSuite.scala
```diff
@@ -0,0 +1,161 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.log
+
+import java.io.File
+
+import com.waz.api.ZmsVersion
+import com.waz.specs.AndroidFreeSpec
+
+class InternalLogSuite extends AndroidFreeSpec {
```
deanrobertcook: Normally we call them `Spec` - so `InternalLogSpec`

zmessaging/src/main/scala/com/waz/log/SystemLogOutput.scala
```diff
@@ -0,0 +1,39 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.log
+
+import com.waz.ZLog.LogTag
+
+import scala.concurrent.Future
+
+class SystemLogOutput extends LogOutput {
```
deanrobertcook: I think this class could go into `testutils`. We don't need it in the main code, do we?

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -0,0 +1,77 @@
...
+
+  override def log(str: String, cause: Throwable, level: InternalLog.LogLevel, tag: LogTag): Unit = this.synchronized {
+    buffer.append(level).append("/").append(tag).append(": ").append(str).append('\n').append(InternalLog.stackTrace(cause))
+    if (size > bufferSize) flush()
+  }
+
+  override def close(): Future[Unit] = flush()
+
+  def empty = buffer.isEmpty
+  def size = buffer.length * 2L // https://stackoverflow.com/a/4387559/2975925
```
deanrobertcook: we need to do a bit more attribution than that to fit with our license - there are some other examples around the app. Checkout the `NetworkModeService`

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -0,0 +1,77 @@
...
+
+import java.io.{BufferedWriter, File, FileWriter, IOException}
+
+import com.waz.ZLog.LogTag
+import com.waz.utils.returning
+
+import scala.concurrent.Future
+
+class BufferedLogOutput(val fileName: String, val bufferSize: Long) extends LogOutput {
+  import scala.concurrent.ExecutionContext.Implicits.global
```
deanrobertcook: use our own `Threading.Implicits.IO`

deanrobertcook: Or even better - we can use a `SerialDispatchQueue` based on `Threading.Implicits.IO`, then we don't need to synchronize anything inside this class

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -0,0 +1,92 @@
...
+import com.waz.threading.{SerialDispatchQueue, Threading}
+import com.waz.utils.returning
+
+import scala.concurrent.Future
+import scala.util.Random
+
+class BufferedLogOutput(val fileName: String, val bufferSize: Long) extends LogOutput {
+  override val id = fileName
+
+  private implicit val dispatcher = new SerialDispatchQueue(Threading.Implicits.BlockingIO, "BufferedLogOutput" + Random.nextInt().toHexString)
```
deanrobertcook: not very important, but here should have been `Threading.IO`. The `.Implicits` object just allows us to import our threading objects implicitly

deanrobertcook: Also, I don't know why we have so many aliases for some of the threading objects

macrosupport/src/main/scala/com/waz/ZLog.scala
```diff
@@ -59,91 +49,73 @@ private object ZLogMacros {
...
     import c.universe._
-    q"""
-        if (com.waz.utils.isTest) {
-          if (com.waz.ZLog.testLogLevel <= com.waz.ZLog.LogLevel.Verbose) ${reify(println(s"V: ${tag.splice}: ${message.splice}"))}
-        }
-        else
-          if (com.waz.ZLog.minimumLogLevel <= android.util.Log.VERBOSE) android.util.Log.v($tag, $message)
-      """
+    DebugMode.DEBUG(c) match {
+      case Literal(Constant(d: Boolean)) if d => q"com.waz.log.InternalLog.verbose($message, $tag)"
```
zbsz: I mean that `$message` is evaluated before `InternalLog.verbose` call, as this is regular function. In old code this operation would be executed only `DEBUG == false`.

PR #164, opened 2017-06-19
[AN-5351][AN-5341] clears memory image cache on low memory

makingthematrix:


zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
```diff
@@ -351,4 +351,17 @@ object ZMessaging { self =>
     }
   }
 
+  // should be called on low memory events
```
zbsz: Take a look at how this is implemented in `TrimmingLruCache`, specifically `AutoTrimming` trait. You could implement it the same way, or even reuse `AutoTrimming` instead of introducing public api and relying on UI to call this method.

zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
```diff
@@ -351,4 +351,17 @@ object ZMessaging { self =>
...
+  def onTrimMemory(level: Int): CancellableFuture[Unit] = level match {
+    case ComponentCallbacks2.TRIM_MEMORY_UI_HIDDEN |
+         ComponentCallbacks2.TRIM_MEMORY_RUNNING_LOW |
+         ComponentCallbacks2.TRIM_MEMORY_RUNNING_CRITICAL =>
+      ExceptionHandler.saveException(new RuntimeException(s"onTrimMemory($level)"), null, null)
+      Threading.Background {
+        currentGlobal.cache.deleteExpired()
```
zbsz: don't think we want to call `deleteExpired` here, this is potentially heavy operation deleting expired items from disk, it will not release any memory, it will actually use memory to perform this operation, I would avoid that

zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
```diff
@@ -351,4 +351,17 @@ object ZMessaging { self =>
...
+        currentGlobal.imageCache.clear()
```
zbsz: this doesn't need to be done in background (if we stop calling `deleteExpired`)
I would suggest evicting in steps based on memory pressure (as in `AutoTrimming`) instead of completely clearing this cache every time. As it is right now it can cause quite a significant amount of additional work when app is resumed, especiallly in group chats where we will need to reload all avatars, chatheads themselves are rather small but sometimes require loading from quite big images

PR #165, opened 2017-06-20
InternalLog creates new files when the max file limit is exceeded and roll them

makingthematrix:


zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -23,50 +23,59 @@ import com.waz.ZLog.LogTag
...
+    if (!empty) {
+      val contents = buffer.toString
+      val path = currentPath
+      if (new File(path).length() + size > maxFileSize) paths = (basePath + paths.size + ".log") :: paths
+      returning(Future {
+        writeToFile(path, contents)
+        if (paths.size > maxRollFiles) {
+          paths = BufferedLogOutput.roll(paths.reverse)
+        }
+      }) { _ => buffer.clear() }
```
zbsz: you could clear buffer before creating a future, wouldn't need `returning` here

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -81,12 +90,28 @@ class BufferedLogOutput(val fileName: String, val bufferSize: Long) extends LogO
...
       }
     } catch {
       case ex: IOException => ex.printStackTrace()
     }
   }
+
+}
+
+object BufferedLogOutput {
+  val defMaxBufferSize = 256L * 1024L
```
zbsz: Scala style guide proposes upper case for constants: `DefMaxBufferSize`

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -81,12 +90,28 @@ class BufferedLogOutput(val fileName: String, val bufferSize: Long) extends LogO
...
+  val defMaxFileSize = 4L * defMaxBufferSize
+  val defMaxRollFiles = 10
+  val defFileName = "internalLog"
+
+  @tailrec
+  private def roll(pathsSorted: List[String], newPaths: List[String] = Nil): List[String] = pathsSorted match {
+    case first :: second :: tail =>
+      new File(second).renameTo(new File(first))
+      if (tail != Nil) roll(second :: tail, first :: newPaths) else first :: newPaths
```
zbsz: no real need to have this `if` here, pattern match in recursive call would handle that

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -23,50 +23,59 @@ import com.waz.ZLog.LogTag
...
   // TODO: In this implementation we risk that writing to the file fails and we lose the contents.
   // But if we wait for the result of writeToFile, we risk that meanwhile someone will add something
   // to the buffer and we will lose that.
   override def flush(): Future[Unit] = this.synchronized {
-    if (!empty) returning(Future { writeToFile(fileName, buffer.toString) }) { _ => buffer.clear() }
-    else Future.successful {}
+    if (!empty) {
+      val contents = buffer.toString
+      val path = currentPath
+      if (new File(path).length() + size > maxFileSize) paths = (basePath + paths.size + ".log") :: paths
```
zbsz: should probably use `while` here, as there might already be several existing files, especially when app is just started and `paths` has only default entry, but disk might have multiple old files, we always want to write to the newest one

this is also not safe for cases when `maxFileSize` is modified, let's say we were using `maxFileSize = 1MB`, we run the app for some time and generate 5 rolled files of this size, then in next update we decide to use `maxFileSize = 2MB` this check will fail and we will start appending logs to the oldest file, and then swith to newer one and append there, basically destroying log ordering

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -23,50 +23,59 @@ import com.waz.ZLog.LogTag
...
+                        val maxRollFiles: Int = BufferedLogOutput.defMaxRollFiles) extends LogOutput {
+  assert(maxBufferSize < maxFileSize)
+  assert(maxRollFiles > 0)
+
+  override val id = basePath
 
   private implicit val dispatcher = new SerialDispatchQueue(Threading.IO, "BufferedLogOutput" + Random.nextInt().toHexString)
 
   private val buffer = StringBuilder.newBuilder
+  private var paths = List[String](basePath + "0.log")
```
zbsz: I would suggest to initialize this list by checking existing files on disk and listing all matching `basePath\d+.log` pattern
this would prevent issues mentioned in `flush` method
make sure to do that on `dispatcher` to ensure this is done before any write operation

zmessaging/src/main/scala/com/waz/log/BufferedLogOutput.scala
```diff
@@ -81,12 +90,28 @@ class BufferedLogOutput(val fileName: String, val bufferSize: Long) extends LogO
...
+}
+
+object BufferedLogOutput {
+  val defMaxBufferSize = 256L * 1024L
+  val defMaxFileSize = 4L * defMaxBufferSize
+  val defMaxRollFiles = 10
+  val defFileName = "internalLog"
+
+  @tailrec
+  private def roll(pathsSorted: List[String], newPaths: List[String] = Nil): List[String] = pathsSorted match {
```
zbsz: this wouldn't handle a case where `maxRollFiles` is changed to something smaller, we would like to delete old files then

PR #174, opened 2017-06-27
checking if the exception is not null before printing the stack trace

makingthematrix:


zmessaging/src/main/scala/com/waz/log/InternalLog.scala
```diff
@@ -70,12 +70,12 @@ object InternalLog {
   def debug(msg: String, tag: LogTag)                   = log(msg, Debug, tag)
   def verbose(msg: String, tag: LogTag)                 = log(msg, Verbose, tag)
 
-  def stackTrace(cause: Throwable) = {
+  def stackTrace(cause: Throwable) = if (cause != null) {
```
deanrobertcook: `Option(cause).foreach { c => ... } `

PR #175, opened 2017-06-29
Removes TeamMembers

makingthematrix:
* `teamId` is now retrieved during logging in and set in `AccountData` and `UserData`
* many methods are simplified or deleted because now there is only one `teamId` (or none) and we know it already
* some changes in UI are required
* no migration yet
* The way teamId is now stored (`Option[TeamId]`) can be improved. Right now `teamId == None` may mean both that we don't know the `teamId` yet and that this is a private account. 
* **more tests are needed...**

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
@@ -162,6 +162,7 @@ class TeamsServiceImpl(selfUser:          UserId,
     verbose(s"onTeamMembersLeft: users: $userIds")
     if (userIds.contains(selfUser)) {
       warn("Self user removed from team")
+      Future.successful {}
     } else {
       for {
         _ <- userStorage.remove(userIds)
--- a/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
@@ -39,15 +39,15 @@ import scala.concurrent.Future.traverse
 //TODO - return Signals of the search results for UI??
 trait TeamsService {
 
-  def getTeam(userId: UserId): Signal[Option[TeamData]]
+  def team(userId: UserId): Signal[Option[TeamData]]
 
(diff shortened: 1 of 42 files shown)


zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -208,48 +153,38 @@ class TeamsServiceImpl(selfUser:          UserId,
 
   //TODO follow up on: https://github.com/wireapp/architecture/issues/13
   //At the moment, we need to re-fetch the entire list of team members as a workaround
-  private def onMembersJoined(teamId: TeamId, users: Set[UserId]) = {
-    verbose(s"onTeamMembersJoined: $teamId, users: $users")
-    sync.syncTeams(Set(teamId))
+  private def onMembersJoined(users: Set[UserId]) = {
+    verbose(s"onTeamMembersJoined: users: $users")
+    sync.syncTeam()
```
deanrobertcook: since we're not storing the `Permissions` in the database anymore, there's no need to sync the team. Here we can just add a user entry with the team id, and then trigger a `syncUser` with the set of ids.

zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -208,48 +153,38 @@ class TeamsServiceImpl(selfUser:          UserId,
...
-  private def removeAllTeamConversations(teams: Set[TeamId]) = {
-    for {
-      convs   <- convsStorage.findByTeams(teams).map(_.map(_.id))
-      _       <- convsStorage.remove(convs)
-      members <- convMemberStorage.getByConvs(convs).map(_.map(m => m.userId -> m.convId))
-      _       <- convMemberStorage.remove(members)
-    } yield {}
-  }
-
   private def removeUnconnectedUsers(users: Set[UserId]): Future[Unit] = {
```
deanrobertcook: I don't think this method is needed anymore either - if a user is removed, and he IS in a team, then that means the user account is gone and we need to delete him anyway (as we already to in `onMembersLeft`) 

I guess the point is, we'll never get `MemberLeave` events for non-connected who aren't also team members users.

That doesn't even make much sense to me - I'll discuss this with you later

PR #183, opened 2017-07-04
Listen to MemberUpdate event and ensure permissions are downloaded and updated

deanrobertcook:


zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
```diff
@@ -100,15 +109,25 @@ object TeamsClient {
...
+          None
+      }
+  }
+
+  object TeamMembersResponse {
+    def unapply(response: ResponseContent): Option[Set[(UserId, PermissionsMasks)]] = {
       response match {
         case JsonObjectResponse(js) if js.has("members") =>
-          Try(decodeSet('members)(js, MemberDecoder)).toOption
+          Try(decodeSet('members)(js, TeamMemberResponse.MemberDecoder)).toOption
```
makingthematrix: Please map the result set to a map: `.Try(decodeSet('members)(js, TeamMemberResponse.MemberDecoder)).toMap.toOption`
and use Map[UserId, PermissionMasks] instead

PR #198, opened 2017-07-20
Make sure we check team id in team member search

hakonbo:


zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -113,11 +113,15 @@ class TeamsServiceImpl(selfUser:           UserId,
       }
     }
 
-    def userMatches(data: UserData) = query match {
-      case Some(q) =>
-        if (handleOnly) data.handle.map(_.string).contains(q.asciiRepresentation)
-        else q.isAtTheStartOfAnyWordIn(data.searchKey)
-      case _ => true
+    def userMatches(data: UserData) = data.teamId match {
```
deanrobertcook: you should just be able to do `data.teamId == teamId && (query.match ...)` - but I also noticed another bug I introduced here - the entire method should just return an `Signal.empty` if the `teamId` of the class is `None`... Maybe you want to add that here too?

PR #199, opened 2017-07-20
[AN-5485] Fix to remove updated values from the search signal current value

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-5485

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
@@ -128,7 +128,7 @@ class TeamsServiceImpl(selfUser:           UserId,
           case Updated(id, _, data) if !userMatches(data) => id
         }.toSet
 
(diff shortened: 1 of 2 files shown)


zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -128,7 +128,7 @@ class TeamsServiceImpl(selfUser:           UserId,
           case Updated(id, _, data) if !userMatches(data) => id
         }.toSet
 
-        current.filterNot(d => removed.contains(d.id)) ++ added
+        current.filterNot(d => removed.contains(d.id) || added.exists(_.id == d.id)) ++ added
```
deanrobertcook: Maybe you can add a test for that case in the `TeamsServiceSpec` - there are already a few examples there on this method

deanrobertcook: And please be sure the other tests are working/are fixed as well

PR #200, opened 2017-07-25
[AN-5406] Searches for the exact handle match

makingthematrix:
* Lots of user searching functionality moved from UI to SE
* If `UserData` with the exact handle match is not found with the standard search, a new request is used to search for that on the backend
* Refactoring to allow android-free unit tests of `UserSearchService` and unit tests covering the standard search
* A hack: garbage collection is enforced before the search to lower the risk of GC suspending the search in-between
* Still to do: more unit tests!

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/sync/SyncRequest.scala
+++ b/zmessaging/src/main/scala/com/waz/model/sync/SyncRequest.scala
@@ -103,6 +103,10 @@ object SyncRequest {
     override val mergeKey: Any = (cmd, query)
   }
 
+  case class ExactMatchHandle(handle: Handle) extends BaseRequest(Cmd.ExactMatchHandle) {
+    override val mergeKey: Any = (cmd, handle)
+  }
+
   case class SyncRichMedia(messageId: MessageId) extends BaseRequest(Cmd.SyncRichMedia) {
     override val mergeKey: Any = (cmd, messageId)
   }
@@ -302,6 +306,7 @@ object SyncRequest {
           case Cmd.SyncUser              => SyncUser(users)
           case Cmd.SyncConversation      => SyncConversation(decodeConvIdSeq('convs).toSet)
           case Cmd.SyncSearchQuery       => SyncSearchQuery(SearchQuery.fromCacheKey(decodeString('queryCacheKey)))
+          case Cmd.ExactMatchHandle      => ExactMatchHandle(Handle(decodeString('handle)))
           case Cmd.PostConv              => PostConv(convId, decodeStringSeq('users).map(UserId(_)), 'name, 'team)
           case Cmd.PostConvName          => PostConvName(convId, 'name)
           case Cmd.PostConvState         => PostConvState(convId, JsonDecoder[ConversationState]('state))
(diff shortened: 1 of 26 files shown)


zmessaging/src/main/scala/com/waz/model/sync/SyncRequest.scala
```diff
@@ -369,6 +374,7 @@ object SyncRequest {
         case SyncUser(users)                  => o.put("users", arrString(users.toSeq map ( _.str)))
         case SyncConversation(convs)          => o.put("convs", arrString(convs.toSeq map (_.str)))
         case SyncSearchQuery(queryCacheKey)   => o.put("queryCacheKey", queryCacheKey.cacheKey)
+        case ExactMatchHandle(handle)              => o.put("handle", handle.string)
```
deanrobertcook: formatting... ;)

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
```diff
@@ -44,30 +44,55 @@ import scala.concurrent.Future
...
+  def createOneToOneConversation(toUser: UserId, selfUserId: UserId): Future[ConversationData]
+  def createGroupConversation(id: ConvId, members: Seq[UserId], teamId: Option[TeamId] = None): Future[ConversationData]
+  def findGroupConversations(prefix: SearchKey, limit: Int, handleOnly: Boolean): Future[Seq[ConversationData]]
+  def knock(id: ConvId): Future[Option[MessageData]]
+  def setLastRead(convId: ConvId, msg: MessageData): Future[Option[ConversationData]]
+  def setEphemeral(id: ConvId, expiration: EphemeralExpiration): Future[Option[(ConversationData, ConversationData)]]
+
+  val assetUploadStarted   : EventStream[AssetData]
+  val assetUploadCancelled : EventStream[Mime]
+  val assetUploadFailed    : EventStream[ErrorResponse]
```
deanrobertcook: for tests these need to be `def`

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
```diff
@@ -378,26 +397,22 @@ class ConversationsUiService(self:            UserId,
...
-        sync.postLastRead(convId, conv.lastRead)
-        Some(conv)
-      case _ => None
-    }
+  override def knock(id: ConvId): Future[Option[MessageData]] = for {
+    msg <- messages.addKnockMessage(id, self)
+    _   <- sync.postMessage(msg.id, id, msg.editTime)
+  } yield Some(msg)
+
+  override def setLastRead(convId: ConvId, msg: MessageData): Future[Option[ConversationData]] = convsContent.updateConversationLastRead(convId, msg.time) map {
```
deanrobertcook: I think it looks a little more readable to drop the start of the method onto the next line

deanrobertcook: (you can leave off the brackets though.)

PR #204, opened 2017-07-25
Unread count and order unchanged with system messages

JPonte:


tests/unit/src/test/scala/com/waz/service/conversation/ConversationOrderEventsServiceSpec.scala
```diff
@@ -85,4 +86,46 @@ class ConversationOrderEventsServiceSpec extends AndroidFreeSpec {
...
+  lazy val storage   = mock[ConversationStorage]
+  lazy val messages  = mock[MessagesService]
+  lazy val users     = mock[UserService]
+  lazy val sync      = mock[SyncServiceHandle]
+
+  scenario("System messages shouldn't change order except it contains self") {
+
+    lazy val scheduler: EventScheduler = new EventScheduler(Stage(Sequential)(service.conversationOrderEventsStage))
+    lazy val pipeline  = new EventPipeline(Vector.empty, scheduler.enqueue)
+    lazy val service = new ConversationOrderEventsService(convs, storage, messages, users, sync, pipeline)
```
deanrobertcook: Generally I think it's safer to provide the mocks with expectations before generating the services, just in case the service needs to do stuff with the mocks in its constructor, but if it works, then fine

PR #205, opened 2017-07-27
[AN-5504] fix; reduces the number of search queries; adds unit tests

makingthematrix:
To be honest, I don't know exactly what caused AN-5504. I focused on optimizations to user search, planning to look into the bug after that. When I finished, the bug was already fixed.

1. In the conversation search SQL query I replaced "FROM ... WHERE ..." with "JOIN". I'm not sure if SQLite doesn't do it anyway, but since we perform raw queries, then maybe not.
2. When searching for a handle in the database, we consider the key to be a prefix, but not a middle part of the handle.
2. TopPeople search is moved out from searchUserData. Since TopPeople is local, there's no need to mix them.
3. In the directory search, when the results are back and we find out that the user already changed the query to something else, we don't put them in the storage, nor update the cache. 
4. Exact match search is triggered by a successful directory search.


makingthematrix:
build again

PR #207, opened 2017-08-02
Simplify downloads

deanrobertcook:
Note - I will still be writing tests for the next day or two - we just want this in internal soon!


zmessaging/src/main/scala/com/waz/service/downloads/AssetLoader.scala
```diff
@@ -0,0 +1,248 @@
...
+}
+
+class AssetLoaderImpl(context:         Context,
+                      network:         NetworkModeService,
+                      client:          AssetClient,
+                      audioTranscoder: AudioTranscoder,
+                      videoTranscoder: VideoTranscoder,
+                      cache:           CacheService) extends AssetLoader {
+
+  private val downloadEnabled = {
```
makingthematrix: This seems like a very complicated way to check user preferences. Why not a for-comprehension?

zmessaging/src/main/scala/com/waz/service/downloads/AssetLoader.scala
```diff
@@ -0,0 +1,248 @@
...
+  override val onDownloadDone     = EventStream[AssetId]()
+  override val onDownloadFailed   = EventStream[(AssetId, ErrorResponse)]()
+
+  override def loadAsset(asset: AssetData, callback: Callback, force: Boolean) = {
+    verbose(s"loadAsset: ${asset.id}, isDownloadable?: ${asset.isDownloadable}, force?: $force, mime: ${asset.mime}")
+    asset match {
+      case _ if asset.mime == Mime.Audio.PCM => transcodeAudio(asset, callback)
+      case _ => CancellableFuture.lift(cache.getEntry(asset.cacheKey)).flatMap {
+        case Some(cd) => CancellableFuture.successful(cd)
+        case None if asset.isDownloadable && (force || downloadEnabled.currentValue.exists(identity)) =>
```
makingthematrix: These two "case None" could be merged
And downloadEnabled.currentValue.exists(identity) could be just a private var

zmessaging/src/main/scala/com/waz/service/downloads/AssetLoaderService.scala
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+    }
+  }
+
+  //Returns a Future, since once actual loading has started, it doesn't make sense to cancel it and waste the work.
+  private def load(entry: LoadEntry): Future[(CacheEntry, Int)] = {
+    val asset = entry.asset
+
+    def load() = entry.loader.loadAsset(entry.asset, entry.state ! _, entry.force)
```
makingthematrix: This is the only place in the code where `entry.loader.loadAsset` is called. Please create a method in `LoadEntry` which will call `loader.loadAsset(asset, state ! _, force)` and then hide the loader.

zmessaging/src/main/scala/com/waz/service/downloads/AssetLoaderService.scala
```diff
@@ -0,0 +1,210 @@
...
+  private def loadEntry(asset: AssetData, force: Boolean = false)(implicit loader: AssetLoader) = {
+    verbose(s"loadEntry(${asset.id}, $force)")
+
+    verbose(s"load requests: ${requests.keys}")
+    verbose(s"active:        $active")
+    verbose(s"in queue:      $queue")
+
+    def createEntry = returning(new LoadEntry(asset, loader)) { onAdded ! _ }
+
+    returning(requests.getOrElseUpdate(asset.id, createEntry)) { entry =>
```
makingthematrix: It seems the only reason `LoadEntry.asset` is a public `var` is because here you modify it just after creating the entry. If you move the `if` outside the `returning` you can turn `asset` into a `val`.

tests/unit/src/test/scala/com/waz/service/assets/AssetLoaderServiceSpec2.scala
```diff
@@ -0,0 +1,121 @@
+/*
```
makingthematrix: Can you merge this with `AssetLoaderServiceSpec`?

zmessaging/src/main/scala/com/waz/service/downloads/AssetLoaderService.scala
deanrobertcook: We could get potentially get another request with the same asset id, but then a different (or updated) `AssetData` that might contain more information - in that case, we don't want create a new `DownloadEntry` with a new time and potentially different properties (or do we...) 

Maybe replacing the Download Entry with a copy would be clearer

tests/unit/src/test/scala/com/waz/service/assets/AssetLoaderServiceSpec2.scala
deanrobertcook: That's still part of my testing to-dos :)

PR #208, opened 2017-08-03
Add more tests for simplified AssetLoaderService

deanrobertcook:
These tests were rewritten to match the new class


tests/utils/src/main/scala/com/waz/specs/AndroidFreeSpec.scala
```diff
@@ -58,8 +57,8 @@ abstract class AndroidFreeSpec extends FeatureSpec with BeforeAndAfterAll with B
 
     ZMessaging.clock = clock
 
-    InternalLog.reset()
-    InternalLog.add(new SystemLogOutput)
+//    InternalLog.reset()
+//    InternalLog.add(new SystemLogOutput)
```
JPonte: Any reason not to just remove?

zmessaging/src/main/scala/com/waz/service/downloads/AssetLoaderService.scala
```diff
@@ -94,45 +97,47 @@ class AssetLoaderService {
 
   private def loadEntry(asset: AssetData, force: Boolean = false)(implicit loader: AssetLoader) = {
     verbose(s"loadEntry(${asset.id}, $force)")
+    verbose(s"load requests: ${requests.keys.map(_.toString).toSeq.sorted}")
+    verbose(s"active:        ${active.map(_.toString).toSeq.sorted}")
+
+    if (!active(asset.id)) {
+      def createOrUpdate(cur: Option[LoadEntry]) = {
```
makingthematrix: I feel that in this case the nested `def` and the following `returning` clause make code less readable instead of more, and a more straightforward approach would be better. But this is just a code-style thing. Leave it as it is if you like it this way.

deanrobertcook: true - it used to have to be a `def` because of the `while` loop - now it's recursive, so I could just use a val

tests/utils/src/main/scala/com/waz/specs/AndroidFreeSpec.scala
deanrobertcook: Because then I have to remember what to enable every time I want logging in tests :)

zmessaging/src/main/scala/com/waz/service/downloads/AssetLoaderService.scala
deanrobertcook: No, actually, I was wrong - I also tried the other way - I think this is a little clearer with the returning

PR #210, opened 2017-08-07
[AN-5476] Specialized sendMessage methods in ConversationsUiService

makingthematrix:
Apart from one general `sendMessage`, `ConversationsUiService` has now a list of specialized versions of the method. They can be used from UI and from other places in SE directly, without the need to go through the old conversation API. Eventually we should be able to replace all uses of the general method with the specialized ones and remove the old API.
To do:
* Tests in `MessageSendingSpec`
* In UI, use the specialized methods in `ConversationFragment`.

### DIFF
--- a/tests/unit/src/test/scala/com/waz/service/conversation/MessageSendingSpec.scala
+++ b/tests/unit/src/test/scala/com/waz/service/conversation/MessageSendingSpec.scala
@@ -49,27 +49,12 @@ import scala.concurrent.duration._
 import scala.util.Random
 import scala.collection.mutable
 
-class MessageSendingSpec extends AndroidFreeSpec with BeforeAndAfter { test =>
+class MessageSendingSpec extends AndroidFreeSpec { test =>
   implicit lazy val dispatcher = Threading.Background
 
   lazy val selfUser = UserData("self user")
-  lazy val user1 = UserData("user 1")
-
-  var messageSync = None: Option[MessageId]
-  var assetSync = None: Option[AssetId]
-  var lastReadSync = None: Option[(ConvId, Instant)]
-
-  before {
-    messageSync = None
-    assetSync = None
-    lastReadSync = None
-  }
-
-  after {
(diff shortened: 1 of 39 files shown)


tests/unit/src/test/scala/com/waz/service/conversation/MessageSendingSpec.scala
```diff
@@ -19,105 +19,114 @@ package com.waz.service.conversation
...
 import org.threeten.bp.Instant
 
-import scala.concurrent.Await
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration._
 import scala.util.Random
+import scala.collection.mutable
 
-class MessageSendingSpec extends FeatureSpec with Matchers with BeforeAndAfter with RobolectricTests with RobolectricUtils { test =>
+class MessageSendingSpec extends AndroidFreeSpec with BeforeAndAfter { test =>
```
deanrobertcook: `AndroidFreeSpec` already extends `BeforeAndAfterEach`, which allows stacking of `before`s and `afters`. It's a little more cumbersome, since you have to overwrite a method rather than just call it, but it ensures that anything we put in `AndroidFreeSpec` gets called too.

zmessaging/src/main/scala/com/waz/api/impl/conversation/BaseConversation.scala
```diff
@@ -116,13 +116,13 @@ abstract class BaseConversation(implicit ui: UiModule) extends IConversation wit
 
   override def getVerified: Verification = data.verified
 
-  def setConversationName(name: String) = conversations.setName(id, name.trim)
+  def setConversationName(name: String) = ??? //conversations.setName(id, name.trim)
```
deanrobertcook: This seems dangerous?

makingthematrix: Sorry!

PR #213, opened 2017-08-11
saves multiple assets from link previews

makingthematrix:
A small refactoring I did while working on AssetLoaderService.
MessageEventProcessor correctly handles link previews with multiple images.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
+++ b/zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
@@ -89,9 +89,11 @@ class MessageEventProcessor(selfUserId:          UserId,
     //ensure we always save the preview to the same id (GenericContent.Asset.unapply always creates new assets and previews))
     def saveAssetAndPreview(asset: AssetData, preview: Option[AssetData]) =
       assets.mergeOrCreateAsset(asset).flatMap {
-        case Some(asset) => preview.fold(Future.successful(Option.empty[AssetData]))(p => assets.mergeOrCreateAsset(p.copy(id = asset.previewId.getOrElse(p.id))))
-        case _ => Future.successful(Option.empty[AssetData])
-      }.map( _.map( Seq(_) ).getOrElse(Seq.empty[AssetData]) )
+        case Some(asset) => preview.fold(Future.successful(Seq.empty[AssetData]))(p =>
+          assets.mergeOrCreateAsset(p.copy(id = asset.previewId.getOrElse(p.id))).map(_.fold(Seq.empty[AssetData])(Seq(_)))
+        )
+        case _ => Future.successful(Seq.empty[AssetData])
+      }
 
     //For assets v3, the RAssetId will be contained in the proto content. For v2, it will be passed along with in the GenericAssetEvent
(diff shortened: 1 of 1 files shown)


zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
```diff
@@ -105,14 +104,13 @@ class MessageEventProcessor(selfUserId:          UserId,
           verbose(s"Received asset v3: $asset with preview: $preview")
           saveAssetAndPreview(asset, preview)
         case (Text(_, _, linkPreviews), _) =>
-          //TODO Dean - handle link previews with multiple images
-          linkPreviews.headOption.fold(Future.successful(Option.empty[AssetData])) {
+          Future.sequence(linkPreviews.map {
             case LinkPreview.WithAsset(a@AssetData.WithRemoteId(_)) =>
               val asset = a.copy(id = AssetId(id.str))
```
JPonte: In case of previews with multiple images, won't this make them all with the same id?

zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
```diff
@@ -121,22 +119,22 @@ class MessageEventProcessor(selfUserId:          UserId,
           saveAssetAndPreview(asset, preview)
         case (ImageAsset(a@AssetData.IsImageWithTag(Preview)), _) =>
           verbose(s"Received image preview for msg: $id. Dropping")
-          Future successful None
+          Future successful Seq.empty[AssetData]
         case (ImageAsset(a@AssetData.IsImageWithTag(Medium)), Some(rId)) =>
           val asset = a.copy(id = AssetId(id.str), remoteId = Some(rId), convId = convId, data = decryptAssetData(a, data))
           verbose(s"Received asset v2 image: $asset")
-          assets.mergeOrCreateAsset(asset)
+          assets.mergeOrCreateAsset(asset).map( _.map( Seq(_) ).getOrElse(Seq.empty[AssetData]) )
```
JPonte: You could also fold to look prettier I guess.

zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
```diff
@@ -87,16 +87,15 @@ class MessageEventProcessor(selfUserId:          UserId,
...
 
     //ensure we always save the preview to the same id (GenericContent.Asset.unapply always creates new assets and previews))
-    def saveAssetAndPreview(asset: AssetData, preview: Option[AssetData]) = {
+    def saveAssetAndPreview(asset: AssetData, preview: Option[AssetData]) =
       assets.mergeOrCreateAsset(asset).flatMap {
         case Some(asset) => preview.fold(Future.successful(Option.empty[AssetData]))(p => assets.mergeOrCreateAsset(p.copy(id = asset.previewId.getOrElse(p.id))))
         case _ => Future.successful(Option.empty[AssetData])
-      }
-    }
+      }.map( _.map( Seq(_) ).getOrElse(Seq.empty[AssetData]) )
```
JPonte: Maybe you could return immediately the Seq instead of returning Option and then mapping to Seq.

PR #216, opened 2017-08-15
Accounts

deanrobertcook:


zmessaging/src/main/scala/com/waz/service/AccountsService.scala
```diff
@@ -303,6 +230,189 @@ class AccountsService(val global: GlobalModule) {
         } yield {}
     }
   }
+
+
+  //New account methods
```
deanrobertcook: Probably don't need this comment

PR #217, opened 2017-08-16
[AN-5406] user search waits for UserData for the exact match before asking for it

makingthematrix:
Also, transliteration may remove the underscore signs from handles, so I introduced an "or" in matching the query with the handle. Either the transliterated or the raw version should match.
And the standard (ie. not exact) handle match is now "startsWith", not "contains". In case of handles, since the query has to start with '@', we don't look for handles which have the query somewhere inside, but only at the beginning.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/api/impl/ConnectionsSearch.scala
+++ b/zmessaging/src/main/scala/com/waz/api/impl/ConnectionsSearch.scala
@@ -32,7 +32,7 @@ class ConnectionsSearch(searchTerm: String, limit: Int, filter: Array[String], a
   private val filteredIds = filter.toSet
   private val query = SearchKey(searchTerm)
   private val predicate: UserData => Boolean = u =>
-    ((query.isAtTheStartOfAnyWordIn(u.searchKey) && !searchByHandleOnly) || u.handle.exists(_.containsQuery(searchTerm)) || (alsoSearchByEmail && u.email.exists(e => searchTerm.trim.equalsIgnoreCase(e.str)))) && ! filteredIds.contains(u.id.str) && (showBlockedUsers || (u.connection != BLOCKED))
+    ((query.isAtTheStartOfAnyWordIn(u.searchKey) && !searchByHandleOnly) || u.handle.exists(_.startsWithQuery(searchTerm)) || (alsoSearchByEmail && u.email.exists(e => searchTerm.trim.equalsIgnoreCase(e.str)))) && ! filteredIds.contains(u.id.str) && (showBlockedUsers || (u.connection != BLOCKED))
   
   private var users = Option.empty[Vector[UserData]]
   
--- a/zmessaging/src/main/scala/com/waz/model/Handle.scala
+++ b/zmessaging/src/main/scala/com/waz/model/Handle.scala
@@ -21,11 +21,11 @@ import java.util.UUID
 
 import com.waz.utils.Locales
 
-case class Handle(string: String) extends AnyVal{
+case class Handle(string: String) extends AnyVal {
   override def toString : String = string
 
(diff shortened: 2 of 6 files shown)


zmessaging/src/main/scala/com/waz/model/Handle.scala
```diff
@@ -24,12 +24,15 @@ import com.waz.utils.Locales
 case class Handle(string: String) extends AnyVal{
   override def toString : String = string
 
-  def containsQuery(query: String): Boolean = {
-    string.contains(Handle.transliterated(Handle.stripSymbol(query)).toLowerCase)
+  // TODO: investigate why Handle.transliterated removes the underscore (_) sign from handles
```
JPonte: Because it uses this id: "Any-Latin; Latin-ASCII; Lower; [^\\ 0-9a-z] Remove"
The ContentSearchQuery contains a transliterator with a different id that doesn't remove underscores.

zmessaging/src/main/scala/com/waz/utils/Locales.scala
```diff
@@ -112,7 +112,7 @@ trait Transliteration {
 }
 
 object Transliteration {
-  private val id = "Any-Latin; Latin-ASCII; Lower; [^\\ 0-9a-z] Remove"
+  private val id = "Any-Latin; Latin-ASCII; Lower; [^\\ 0-9a-z_] Remove"
```
JPonte: This will make the SearchKeys also contain underscores, which they previously didn't and probably shouldn't...

zmessaging/src/main/scala/com/waz/api/impl/ConnectionsSearch.scala
```diff
@@ -32,7 +32,7 @@ class ConnectionsSearch(searchTerm: String, limit: Int, filter: Array[String], a
   private val filteredIds = filter.toSet
   private val query = SearchKey(searchTerm)
   private val predicate: UserData => Boolean = u =>
-    ((query.isAtTheStartOfAnyWordIn(u.searchKey) && !searchByHandleOnly) || u.handle.exists(_.containsQuery(searchTerm)) || (alsoSearchByEmail && u.email.exists(e => searchTerm.trim.equalsIgnoreCase(e.str)))) && ! filteredIds.contains(u.id.str) && (showBlockedUsers || (u.connection != BLOCKED))
+    ((query.isAtTheStartOfAnyWordIn(u.searchKey) && !searchByHandleOnly) || u.handle.exists(_.startsWithQuery(searchTerm)) || (alsoSearchByEmail && u.email.exists(e => searchTerm.trim.equalsIgnoreCase(e.str)))) && ! filteredIds.contains(u.id.str) && (showBlockedUsers || (u.connection != BLOCKED))
```
JPonte: If the transliterators are the same, the result of startsWithQuery(searchTerm) will be the same as startsWith(searchKey) right?

zmessaging/src/main/scala/com/waz/utils/Locales.scala
makingthematrix: OK. I restored the previous version. When comparing the handle we simply don't use transilteration, assuming that it contains only Latin characters.

zmessaging/src/main/scala/com/waz/api/impl/ConnectionsSearch.scala
makingthematrix: `startsWithQuery` also removes the `@` sign from the beginning if necessary.

JPonte: So did the transliteration, it removed all characters except underscores. But since you're removing it, it makes more sense now.

PR #218, opened 2017-08-16
[AN-5241] images restart downloading when switching network

makingthematrix:


zmessaging/src/main/scala/com/waz/ui/UiModule.scala
```diff
@@ -138,6 +138,7 @@ class UiModule(val accounts: AccountsService) extends UiEventContext with ZMessa
   lazy val assets = new UiCache[AssetId, Asset](10)(this)
 
   lazy val globalImageLoader = global.imageLoader
+  lazy val globalNetwork = global.network
```
deanrobertcook: `network` should be enough - there are no non-global network services.

deanrobertcook:
Also - as we discussed on Friday - look at removing the `vars` and only setting the listener on `DownloadFailedWifiException` (or whatever it's called)

tests/unit/src/test/scala/com/waz/service/images/BitmapSignalSpec.scala
```diff
@@ -17,128 +17,192 @@
...
+      (loader.loadBitmap _).expects(expectedAsset, req, *).once().onCall { _ =>
+        val bitmap = mockBitmap(expectedAsset)
+        bmps ! Some(bitmap)
+        CancellableFuture.successful(bitmap)
+      }
+
+      (loader.loadCachedBitmap _).expects(expectedAsset, req).anyNumberOfTimes.onCall { _ =>
+        bmps.currentValue.flatten match {
+          case Some(bitmap) => CancellableFuture.successful(bitmap)
+          case None => CancellableFuture.failed(new IllegalStateException("Trying to return a non-existing cached bitmap"))
```
deanrobertcook: if this should fail the test, you can also just call `fail("some failure message")`

zmessaging/src/main/scala/com/waz/service/images/BitmapSignal.scala
```diff
@@ -20,60 +20,63 @@ package com.waz.service.images
...
-        loader.load() flatMap { loader.process(_, signal) }
+    future = loader.loadCached().flatMap {
+      case Some(data) => loader.process(data, signal)
+      case None => loader.load().flatMap { loader.process(_, signal) } // will try with download
     }
     future onFailure {
-      case _: CancelException => // ignore
+      case _: CancelException =>
+      case ex@AssetLoader.DownloadOnWifiOnlyException =>
+        signal publish LoadingFailed(ex)
```
deanrobertcook: you can just use `signal ! LoadingFailed(ex)`

deanrobertcook:
merge whenever your happy, and then please create a release with the appropriate fixes :)

zmessaging/src/main/scala/com/waz/service/images/BitmapSignal.scala
makingthematrix: I can't. BitmapSignal is not a SourceSignal. I can use publish, but the "!" shortcut is not defined. (kinda weird)

PR #228, opened 2017-08-30
Notifications for multiple accounts

deanrobertcook:


zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -24,57 +24,114 @@ import com.waz.ZLog._
...
+
+  import com.waz.threading.Threading.Implicits.Background
+
+  lazy val groupedNotifications: Signal[Seq[(AccountId, Boolean, Seq[NotificationInfo])]] = //Boolean = shouldBeSilent
+    Option(ZMessaging.currentAccounts) match {
+      case Some(accountsService) =>
+        accountsService.zmsInstances.flatMap { zs =>
+          val services = zs.map(z => z.accountId -> z.notifications).toSeq.sortBy { case (id, _) => id.str }
+          Signal.sequence(services.map { case (id, s) =>
+            val notifications = s.notifications
```
makingthematrix: How about movingthe part after the `case (id, s)` into a separate method? Would improve readability.

deanrobertcook: I disagree - I often try and remove such small, one-time-use methods because they run the risk of being (physically) separated from the point they're being used in the code, and then you have to come up with new variable names and annotate the types - all of which is more code and requires you to hold more in your head while reading it :)

PR #232, opened 2017-08-31
postpones syncing after PushService is done processing

makingthematrix:
Prevents one possible reason for CryptoBox detecting outdated messages.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
@@ -72,7 +72,9 @@ class ConversationsService(context:         Context,
     verbose(s"onSlowSyncNeeded($req)")
     for {
       _ <- if (req.lostHistory) onHistoryLost() else successful(())
-      _ <- sync.syncConversations()
+      processing <- push.processing.head
+      if !processing
+        _ <- sync.syncConversations()
       _ <- sync.syncTeam()
     } yield ()
   }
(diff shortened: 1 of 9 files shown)


zmessaging/src/main/scala/com/waz/service/push/PushServiceImpl.scala
```diff
@@ -195,6 +195,9 @@ object PushService {
 class PushServiceSignals {
   val onSlowSyncNeeded = new SourceSignal[SlowSyncRequest] with BgEventSource
   val pushConnected = Signal[Boolean]()
+  val processing = Signal(false)
+
+  def afterProcessing[T](f : => Future[T])(implicit ec: ExecutionContext): Future[T] = processing.filter(_ == false).head.flatMap(_ => f)
```
deanrobertcook: It might be a bit nicer to move these into the `trait` and implementation of `PushService` and pass the whole service around as a dependency - I feel as though this was to simplify old tests.

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
```diff
@@ -72,7 +72,9 @@ class ConversationsService(context:         Context,
     verbose(s"onSlowSyncNeeded($req)")
     for {
       _ <- if (req.lostHistory) onHistoryLost() else successful(())
-      _ <- sync.syncConversations()
+      processing <- push.processing.head
+      if !processing
+        _ <- sync.syncConversations()
```
deanrobertcook: I don't think slow sync needs to wait on whether or not we're processing notifications - there's no decryption involved here, and normally we only detect that we need to do a slow sync after all notifications have been processed anyway (at least, we should!)

PR #235, opened 2017-09-05
[AN-5486] [AN-5499] PushService reworked in order to avoid race conditions

makingthematrix:
* `LastNotificationIdService` and `PushSignals` merged into `PushService`
* `EventsClient` renamed to `PushNotificationsClient` and simplified - there's no need now to query for `lastNotificationId`
* `PushService` is now getting results from `PushNotificationsClient` directly, not through a signal, which should prevent race conditions.
* `lastNotificationId` updated from the list of retrieved notifications
* Initial sync moved to `AccountManager`
* web socket notifications received during a fetch are processed only after the fetch or dismissed if the fetch fails
* a failed fetch is retried both after a delay and when the network changes


zmessaging/src/main/scala/com/waz/service/AccountManager.scala
```diff
@@ -184,6 +185,12 @@ class AccountManager(val id: AccountId, val global: GlobalModule, accounts: Acco
   // logged in zmessaging instance
   @volatile private var _zmessaging = Option.empty[ZMessaging]
 
+  private val shouldSyncInitial = storage.userPrefs.preference(ShouldSyncInitial)
+  for {
+    um         <- userModule
+    shouldSync <- shouldSyncInitial.signal
+  } if (shouldSync) um.sync.performFullSync().flatMap(_ => shouldSyncInitial := false)
```
deanrobertcook: I thought of a slight issue with this actually - if we have a `sync_initial` value for everything (users, convs, teams, clients etc) and use it to indicate we should sync everything - then we may have issues in the future if we decide to add another entity, which will also need a full-sync after updating, but we don't necessarily need to sync everything else - maybe not a huge deal though.

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -0,0 +1,243 @@
...
+/** PushService handles notifications coming from FCM, WebSocket, and fetch.
+  * We assume FCM notifications are unreliable, so we use them only as information that we should perform a fetch (syncHistory).
+  * A notification from the web socket may trigger a fetch on error. When fetching we ask BE to send us all notifications since
+  * a given lastId - it may take time and we even may find out that we lost some history (lastId not found) which triggers slow sync.
+  * So, it may happen that during the fetch new notifications will arrive through the web socket. Their processing should be
+  * performed only after the fetch is done. For that we use a serial dispatch queue and we process notifications in futures,
+  * which are put at the end of the queue, essentially making PushService synchronous.
+  * Also, we need to handle fetch failures. If a fetch fails, we want to repeat it - after some time, or on network change.
+  * In such case, new web socket notifications waiting to be processed should be dismissed. The new fetch will (if successful)
+  * will receive them in the right order.
```
deanrobertcook: great

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -0,0 +1,243 @@
...
+    case _ => Signal.empty[Set[Uid]]
+  }.on(dispatcher) {
+    notifications =>
+      if (notifications.nonEmpty && fetchInProgress.isCompleted){
+        verbose("Sync history in response to cloud message notification")
+        syncHistory()
+      }
+  }
+
+  override def onPushNotification(n: PushNotification) = onPushNotifications(Seq(n)) //used in tests
```
deanrobertcook: I think we can get rid of this method?

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -0,0 +1,243 @@
...
+    case Left(err) =>
+      warn(s"Request failed due to $err: attempting to load last page (since id: $lastId) again")
+      CancellableFuture.delay(backoff.delay(attempts)).flatMap { _ => loadNotifications(lastId, attempts + 1) }
+      // fetch failed, schedule retry and turn on retry on network change
+  }
+
+  private lazy val idPref = userPrefs.preference(LastStableNotification)
+  def lastNotificationId: Future[Option[Uid]] = { Future(idPref()).flatten } // future and flatten ensure that there is no race with onNotification
+  def updateLastNotificationId(id: Uid): Future[Unit] = { idPref := Some(id) }
+  def resetLastNotificationId(): Future[Unit] = { idPref := None }
```
deanrobertcook: I think these 3 methods are unnecessary - especially now that everything is synchronous

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -0,0 +1,243 @@
...
+    case _ =>
+  }
+
+  override def syncHistory(): Future[Unit] = returning( lastNotificationId flatMap { syncHistory } ) { f =>
+    if (!fetchInProgress.isCompleted) fetchInProgress.cancel()
+      // a new fetch was ordered so we cancel the old one if it's still in progress, together with any ws notifications processing attached to it
+    fetchInProgress = CancellableFuture.lift(f)
+  }
+
+  private def syncHistory(lastId: Option[Uid]): Future[Unit] =
```
deanrobertcook: It might be a bit nicer to have this as an internal method in the other `syncHistory()`

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -0,0 +1,243 @@
...
+    notifications =>
+      if (notifications.nonEmpty && fetchInProgress.isCompleted){
+        verbose("Sync history in response to cloud message notification")
+        syncHistory()
+      }
+  }
+
+  override def onPushNotification(n: PushNotification) = onPushNotifications(Seq(n)) //used in tests
+
+  override def onPushNotifications(allNs: Seq[PushNotification]): Unit = if (allNs.nonEmpty) {
```
deanrobertcook: I think this method can actually be private, and merged together with `processNotifications`

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,460 @@ package com.waz.service.push
...
+  feature("Updating lastNotificationId pref") {
 
-  def lastNotificationId_=(id: Option[Uid]) = Await.result(service.lastNotification.idPref := id, 5.seconds)
+    scenario("Update pref on fresh notification") {
+      waitForLastNotIdChangeTo(None)
+
+      service.updateLastNotificationId(Uid(3, 3))
+      service.updateLastNotificationId(Uid(4, 4))
+      service.updateLastNotificationId(Uid(5, 5))
+      waitForLastNotIdChangeTo(Some(Uid(5, 5)))
```
deanrobertcook: All that you're really testing here is that `TestPreferences` works - `updateLastNotificationId` is really an internal part of the class (and as I mention above, doesn't need its own method). 

Instead of just pushing to the preferences, you should mock the `webSocketClientService` to return you mocked versions of `WebSocketClient` (when the `webSocket.client` signal returns you one) and then push a fake message down that signal. That would then test as much of the internals of the class as possible - testing the chaining of futures and that the correct dependencies are called when processing a notification.

You could also perform the same test after a `syncHistory`, and then a combination of the two

zmessaging/src/main/scala/com/waz/service/AccountManager.scala
makingthematrix: OK. Noted. I leave it for now.

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
makingthematrix: done

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
makingthematrix: done

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
makingthematrix: done

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: Removed and replaced with one test to check if a push from cloud is setting `lastNotificationId`. Checking if this happens for web socket and fetch is already handled by other tests.

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
+      |                  "nonce":"47745f9f0-0dab-113c-43ad7ee9-394c562"
+      |               },
+      |               "from":"13962457-c316-4de1-9962-929c40f8cff4",
+      |               "id":"f.80011231430865a7",
+      |               "time":"2014-04-14T09:56:00.185Z",
+      |               "type":"conversation.message-add"
+      |            }
+      |         ]
+      |      }
+    """.stripMargin
```
deanrobertcook: I'm not sure about using the json here directly, tbh. I mean, it's kind of nice to see what we think we're getting from the backend, but then it gets out of date very quickly and won't become a compiler error if it changes (like if we change the interface to `PushNotification`. 

I think it's probably a bit nicer just to create a `PushNotification` directly and use that. Parsing the JSON should perhaps be a test of the `PushNotification#NotificationDecoder` or the `PushNotificationsClient`.

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
+
+      result(wsConnected.filter(_ == false).head)
+
+      cloudPush ! Set(pushNotification.id)
+
+      waitForLastNotIdChangeTo(Some(pushNotification.id))
+    }
+
+    scenario("fetch last notification when there is no local last notification id") {
+      updateLastNotificationId(lastId)
```
deanrobertcook: I'm confused here: `when there is no local last notification id`? But then you set a `lastId`?

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
-    }
+  (wscService.connected _).expects().anyNumberOfTimes().returning(wsConnected)
+  (wscService.client _).expects().anyNumberOfTimes().returning(wsClient)
+  (wscService.network _).expects().anyNumberOfTimes().returning(network)
+  (pipeline.apply _).expects(*).anyNumberOfTimes().returning(Future.successful({}))
 
-    override lazy val websocket = new WebSocketClientService(context, accountId, lifecycle, zNetClient, auth, network, global.backend, clientId, timeouts, pushToken) {
-      override val connected = wsConnected
-    }
+  var syncPerformed = 0
```
deanrobertcook: Here would be better to use a `Signal(0)`, which gets pushed to by the mocked `performFullSync` method. Then you can do `syncPerformed.filter(_ == 1).head` later on

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
-      withDelay {
-        lastNotificationId should be('defined)
-      }
-      service.onPushNotification(PushNotification(Uid(), Nil, transient = true))
-      awaitUi(1.second)
-      lastNotificationId shouldEqual Some(lastId)
+      result(wsConnected.filter(_ == true).head)
+
+      waitForLastNotIdChangeTo(Some(notification2.id))
+      syncPerformed shouldEqual 0
```
deanrobertcook: Here is a good example of where `awaitAllTasks` should be used - or else you're not really testing that the value isn't touched. The test will end on it's thread, and potentially in the future (if there is regression we want to catch) a `sync` could be performed, but we've already checked that it's 0, so it won't fail

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
-        lastNotificationId should be('defined)
-      }
-      service.onPushNotification(PushNotification(Uid(), Seq(OtrMessageEvent(RConvId(), new Date, UserId(), ClientId(), ClientId(), Array.empty))))
-      awaitUi(1.second)
-      lastNotificationId shouldEqual Some(lastId)
+      result(wsConnected.filter(_ == true).head)
+
+      result(service.onHistoryLost.filter(_ == time).head)
+      syncPerformed shouldEqual 1
+      // TODO: this test reports an NPE to Hockey; probably one of the futures finishes after the test has finished
```
deanrobertcook: using the signal suggested above should hopefully fix this

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
-      withDelay {
-        slowSyncRequested shouldEqual 1
-        lastNotificationId shouldEqual Some(lastId)
+      val notification1 = PushNotification(Uid(), Nil)
+      val notification2 = PushNotification(Uid(), Nil)
+      (client.loadNotifications _).expects(*, *).once().returning(
+        CancellableFuture.successful(Right( LoadNotificationsResponse(Vector(notification1, notification2), hasMore = false, None) ))
+      )
+
+      val backoff = mock[Backoff]
```
deanrobertcook: there is a `TestBackoff` class

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -0,0 +1,230 @@
...
+    returning( Future(idPref()).flatten flatMap { syncHistory } ) { f => // future and flatten ensure that there is no race with onPushNotifications
+      if (!fetchInProgress.isCompleted) fetchInProgress.cancel()
+      // a new fetch was ordered so we cancel the old one if it's still in progress, together with any ws notifications processing attached to it
+      fetchInProgress = CancellableFuture.lift(f)
+    }
+  }
+}
+
+object PushService {
+  private var syncHistoryBackoff: Backoff = new ExponentialBackoff(3.second, 15.seconds)
```
deanrobertcook: I think a public var should be fine, it's not like we have that many users of the API who might potentially modify this.

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
+      )
 
-    scenario("fetch last notification when there is no local last notification id") {
-      wsConnected ! true
-      withDelay {
-        slowSyncRequested shouldEqual 1
-        lastNotificationId shouldEqual Some(lastId)
+      val notification1 = PushNotification(Uid(), Nil)
+      val notification2 = PushNotification(Uid(), Nil)
+      (client.loadNotifications _).expects(*, *).once().returning(
```
deanrobertcook: I'm pretty sure mocking the same behaviour a second time wipes the first expectation - instead you should use `onCall` and keep a count of how many times the method was called, matching on it to provide different results. See the `AssetLoaderSpec` for an example

I think if you follow the logs, this test won't actually cause a failure, but since you specified that it should receive both notifications in the second call, it does exactly that and then updates the id as expected

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: simplified and moved below - next to the only test which uses it

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
+    override lazy val wakeLock: WakeLock = new FakeWakeLock
+
+    override def connect(): CancellableFuture[WebSocket] = CancellableFuture.successful(mock[WebSocket])
   }
 
-  after {
-    wsConnected ! false
-    awaitUi(50.millis)
-  }
+  val lastId = Uid()
```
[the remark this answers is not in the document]
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
 
-  def lastNotificationId = Await.result(service.lastNotification.lastNotificationId(), 5.seconds)
+  private lazy val idPref = userPrefs.preference(LastStableNotification)
+  def updateLastNotificationId(id: Uid): Future[Unit] = { idPref := Some(id) }
```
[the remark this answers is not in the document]
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
+  private def waitForLastNotIdChangeTo(id: Option[Uid]) = result(idPref.signal.filter(_ == id).head)
```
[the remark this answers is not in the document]
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: Right, `lastNotificationId` should be `None` and it should set at the init. Changed to `result(idPref.signal.filter(_.isEmpty).head)` (without setting anything) in order to test it.

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
```diff
@@ -19,185 +19,416 @@ package com.waz.service.push
...
+
+      waitForLastNotIdChangeTo(Some(notification2.id))
+      syncPerformed shouldEqual 0
     }
 
-    scenario("don't update id on otr notification not intended for us") {
-      lastNotificationId = None
+    scenario("request slow sync and fetch last notification if /notifications returns error") {
+      AndroidFreeSpec.clock.reset()
+      AndroidFreeSpec.clock.advance(10 minutes)
```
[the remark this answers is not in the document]
makingthematrix: ok, changed

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: done

tests/unit/src/test/scala/com/waz/service/push/PushServiceSpec.scala
makingthematrix: done

zmessaging/src/main/scala/com/waz/utils/WakeLock.scala
```diff
@@ -76,3 +81,8 @@ class TimedWakeLock(context: Context, duration: FiniteDuration)(implicit tag: Lo
 
   private def releaseAfterDuration()(implicit srcTag: LogTag): Unit = CancellableFuture.delay(duration).onComplete(_ => release()(s"delayed[$srcTag]"))
 }
+
+class FakeWakeLock extends WakeLock {
```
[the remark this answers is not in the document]
makingthematrix: done

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
makingthematrix: done

PR #246, opened 2017-09-20
Ensure that chained futures are performed atomically in CallingService

deanrobertcook:


tests/unit/src/test/scala/com/waz/service/call/CallingServiceSpec.scala
```diff
@@ -488,6 +490,41 @@ class CallingServiceSpec extends AndroidFreeSpec {
...
+
+
+      val b = Future {
+        println("1B")
+        Thread.sleep(1000)
+        2
+      } (Threading.Background)
+
+
+      Future (println(s"Result of operations: ${result(a) + result(b)}"))
```
makingthematrix: You should make it `result(a) - result(b)`, so if the order is wrong, the result of the computations is negative :)

PR #247, opened 2017-09-21
Make MediaManager global and tidy up access to internal values

deanrobertcook:


zmessaging/src/main/scala/com/waz/service/MediaManagerService.scala
```diff
@@ -21,56 +21,73 @@ import android.content.Context
...
+  mediaManager.onFailure {
+    case NonFatal(e) =>
+      error("MediaManager was not instantiated properly", e)
   }
 
-  lazy val soundIntensity = prefs.preference(Sounds).signal
+  val isSpeakerOn = RefreshingSignal(mediaManager.map(_.isLoudSpeakerOn), onPlaybackRouteChanged)
 
-  soundIntensity { intensity => withMedia { _.setIntensity(intensity) } }
+  val soundIntensity = Option(ZMessaging.currentAccounts).map(_.activeZms).map {
```
makingthematrix: You can shorten it a bit to:
```
val soundIntensity = Option(ZMessaging.currentAccounts).map(_.activeZms.flatMap {
    case None => Signal.const(IntensityLevelCodec.default)
    case Some(z) => z.userPrefs.preference(Sounds).signal
  }).getOrElse {
    warn("No CurrentAccounts available - this may be being called too early...")
    Signal.const(IntensityLevelCodec.default)
  }
```

PR #249, opened 2017-09-24
Remove unused spotify dependencies.

### LINKED ISSUE #202: Is spotifyPlayer still needed?
As of [wire-android#805](https://github.com/wireapp/wire-android/pull/805) the spotifyPlayer dependency was removed from wire-android. Is it possible to also remove it in this component? This is the main blocker for a [free build](https://github.com/wireapp/wire-android/issues/233).

wire-android:
fixes #202


wire-android:
Turns out we still use those, as we still have player api classes on engine side, not sure if we really want to remove them (may wan to use it in future versions).

PR #250, opened 2017-09-24
Add check for unsupported devices in ClientWrapper

wire-android:
Turns out that a hack for SNI extensions in ClientWrapper doesn't always work.
It was causing crashes:
https://rink.hockeyapp.net/manage/apps/35954/app_versions/517/crash_reasons/183746287

Added check for that case to avoid crashes, kept logging for it to keep stats
about how often this happens and on what devices. We should look into those
devices.


wire-android:
Also added missing call to `setAccessible` which was causing some other error reports.

deanrobertcook:
This will be unnecessary due to https://github.com/wireapp/wire-android-sync-engine/pull/237

PR #251, opened 2017-09-26
Use boolean instead of int in internal methods for cbr, and reenable it

deanrobertcook:


deanrobertcook:
Will wait for new AVS API before merging

PR #252, opened 2017-09-29
[AN-5436] Unread message count fix

JPonte:


wire-android:
There should be no need for that, as lastRead time should be updated every time self message is received. At least it was supposed to be like that some time ago, the idea was that if you receive self message that means everything before it was most likely read on other device already.

JPonte:
But there seems to be an issue with larger assets. The last read is set to the time the message is sent, but it might upload when the user is no longer in the conversation. If there are others messages while it's uploading, when it finishes the timestamp is updated but we can't set the last read to that one because we are no longer in the conversation.

PR #254, opened 2017-10-05
Some notification improvements

hakonbo:


zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -310,11 +313,16 @@ class NotificationService(context:         Context,
...
               msg  <- data.referencedMessage.fold2(Future.successful(None), messages.getMessage)
-              conv <- convs.get(data.conv)
+              conv <- convs.get(data.conv) if conv.isDefined
+              membersCount <- members.getByConv(conv.get.id).map(_.map(_.userId).size)
             } yield {
               val (g, t) =
                 if (data.msgType == LIKE) (data.copy(msg = msg.fold("")(_.contentString)), msg.map(m => if (m.msgType == Message.Type.ASSET) LikedContent.PICTURE else LikedContent.TEXT_OR_URL))
                 else (data, None)
+
+              val groupConv = if (!conv.exists(_.team.isDefined)) conv.exists(_.convType == ConversationType.Group)
```
JPonte: Seems to be unused

hakonbo: Nice catch, should be used, I'll update.

PR #277, opened 2017-11-30
[AN-5721] Store notifications before decrypting

CalumMcCall:
We now store all notifications upon receipt, then event handlers decrypt them in place, and finally parse them from JSON and hand them over to the pipeline.

I also removed OtrAssetEvent as we don't use that message type any more.


CalumMcCall:
This PR is now ready for review. I've tested basic functionality, like sending/receiving messages. I'm currently doing more thorough testing.

zmessaging/src/main/scala/com/waz/service/otr/OtrServiceImpl.scala
```diff
@@ -68,58 +69,48 @@ class OtrServiceImpl(selfUserId: UserId, clientId: ClientId, val clients: OtrCli
...
 
-  def eventTransformer(events: Vector[Event]): Future[Vector[Event]] = {
-    @tailrec def transform(es: Vector[Event], accu: Vector[Future[Vector[Event]]]): Vector[Future[Vector[Event]]] =
-      if (es.isEmpty) accu
-      else {
-        val ph = isOtrEvent(es.head)
-        val (batch, remaining) = es.span(isOtrEvent(_) == ph)
-        val batched = if (ph) collectEvents(clientId, batch) else Future.successful(batch)
-        transform(remaining, accu :+ batched)
+  override def filterOtrEventsForOtherClients(events: JSONArray): JSONArray = {
```
deanrobertcook: Could be a private method or nested function in `PushNotificationEncodedStoarge#insert`: the values for `ClientId`, `UserId` and `AccountId` are essentially global and static from the perspective of any class instantiated in `ZMessaging`, so don't worry if you have to inject them

zmessaging/src/main/scala/com/waz/service/push/PushNotificationEncodedStorage.scala
```diff
@@ -0,0 +1,123 @@
...
+
+  private def insert(notification: PushNotificationEncoded)
+                    (implicit ec: ExecutionContext): Future[Unit] = {
+    import com.waz.model.JSONHelpers._
+    rowStorage.get(notification.id).flatMap {
+      case Some(_) =>
+        sys.error(s"PushNotification with id ${notification.id} already exists, ignoring")
+      case None =>
+        val filtered = otrService.filterOtrEventsForOtherClients(notification.events)
+        if(filtered.length() != 0) {
```
deanrobertcook: slightly more idiomatic may be `filtered.toVector.nonEmpty` (or `JSONArrayToVector`)

zmessaging/src/main/scala/com/waz/service/push/PushNotificationEncodedStorage.scala
```diff
@@ -0,0 +1,123 @@
...
+                sys.error(s"Failed to insert PushNotification into DB: ${e.toString}")
+            }
+            .map(_ => Unit)
+        } else {
+          Future.successful(())
+        }
+    }
+  }
+
+  private def removeEventsWithId(id: Uid): Future[Unit] =
```
deanrobertcook: I try and avoid splitting up a method for the sake of it. Generally scala is compact enough that having a nested def or just inlining the logic altogether is still very readable and prevents the helper methods from becoming to far separated from where they are used - requiring a lot of back-and-forth reading. This is more of an issue in big classes though where fragmentation of the logic, I find, really hurts readability

zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -102,6 +107,41 @@ class PushServiceImpl(context:        Context,
 
   private var subs = Seq.empty[Subscription]
 
+  notificationStorage.processStoredEvents(processEncryptedNotifications)
+  notificationStorage.registerEventsAddedHandler(processEncryptedNotifications)
```
deanrobertcook: you can probably just lump these together into one method, where when you register the `EventsAddedHandler`, we first load any stored notifications and process them. Should then reduce the risk of anyone refactoring this and forgetting to do both.

deanrobertcook: then you can just inline `processEncryptedNotificaitons

zmessaging/src/main/scala/com/waz/model/Event.scala
```diff
@@ -255,6 +270,109 @@ object ConversationEvent {
...
+            .setType("otr-error.identity-changed-error")
+        case Duplicate => json.setType("otr-error.duplicate")
+        case e => throw new JSONException(s"Encoder for event $e not implemented")
+      }
+    }
+  }
+}
+
+object JSONHelpers {
+  implicit class RichJSONArray(js: JSONArray) {
```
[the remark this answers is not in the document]
CalumMcCall: Ah ofc, that is a lot nicer, there is even a RichJSON object already :)

zmessaging/src/main/scala/com/waz/model/Event.scala
```diff
@@ -255,6 +270,109 @@ object ConversationEvent {
...
+    def foldLeft[B](z: B)(op: (B, JSONObject) => B): B = {
+      var result = z
+      for (i <- 0 until js.length()) {
+        result = op(result, js.getJSONObject(i))
+      }
+      result
+    }
+
+    def JSONArrayToVector: Seq[JSONObject] =
```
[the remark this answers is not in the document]
CalumMcCall: Yeah good shout, that's a better name

zmessaging/src/main/scala/com/waz/service/otr/OtrServiceImpl.scala
CalumMcCall: Ah ok that makes more sense aye.

zmessaging/src/main/scala/com/waz/service/push/PushNotificationEncodedStorage.scala
CalumMcCall: Good catch, this has actually made this method much more readable.

zmessaging/src/main/scala/com/waz/service/push/PushNotificationEventsStorage.scala
```diff
@@ -0,0 +1,134 @@
...
+            .zipWithIndex
+            .map { case (obj, index) =>
+              PushNotificationEvent(pn.id, index, decrypted = false, obj, transient = pn.transient)
+            }
+        } else {
+          warn(s"Got notification ${pn.id} containing no events for us, ignoring")
+          Seq.empty[PushNotificationEvent]
+        }
+      }
+    verbose(s"Pipeline: Inserted ${eventsToSave.size} events into db")
```
deanrobertcook: Remove `Pipeline` sub tags

zmessaging/src/main/scala/com/waz/service/push/PushNotificationEventsStorage.scala
```diff
@@ -0,0 +1,134 @@
...
+  def encryptedEvents: Future[Seq[PushNotificationEvent]]
+  def removeEventsWithIds(ids: Seq[Uid]): Future[Unit]
+  def registerEventHandler(handler: Subscriber[Seq[PushNotificationEvent]])(implicit ec: EventContext): Future[Unit]
+}
+
+class PushNotificationEventsStorageImpl(context: Context, storage: Database, clientId: ClientId)
+  extends CachedStorageImpl[(Uid, EventIndex), PushNotificationEvent](new TrimmingLruCache(context, Fixed(100)), storage)(PushNotificationEventsDao, "PushNotificationEvents_Cached")
+    with PushNotificationEventsStorage {
+
+  implicit val logTag: LogTag = logTagFor[PushNotificationEventsStorageImpl]
```
deanrobertcook: So, if you don't care about having a logtag with an account Id, it's slightly nicer to import `com.waz.ZLog.ImplicitTag._` at the top of the file. Otherwise use the `ZMessaging.accountTag` method to get a logtag that includes the first 8 letters of the account id

zmessaging/src/main/scala/com/waz/service/otr/OtrServiceImpl.scala
```diff
@@ -42,32 +41,35 @@ import com.waz.utils.{LoggedTry, _}
...
+  def parseGenericMessage(msgEvent: OtrMessageEvent, msg: GenericMessage): Option[MessageEvent]
 }
 
-class OtrServiceImpl(selfUserId: UserId, clientId: ClientId, val clients: OtrClientsService, push: PushService,
-                 cryptoBox: CryptoBoxService, members: MembersStorageImpl, convs: ConversationsContentUpdaterImpl,
-                 sync: SyncServiceHandle, cache: CacheService, metadata: MetaDataService, clientsStorage : OtrClientsStorage,
+class OtrServiceImpl(selfUserId: UserId, clientId: ClientId, val clients: OtrClientsService,
+                 cryptoBox: CryptoBoxService, members: MembersStorageImpl, sync: SyncServiceHandle,
+                 cache: CacheService, metadata: MetaDataService, clientsStorage : OtrClientsStorage,
                  prefs: GlobalPreferences, tracking: TrackingService) extends OtrService {
```
deanrobertcook: yikes, in these trait definitions I'm happy for the method lines to extend out a bit past the width limit. Otherwise if it gets out of hand, have every argument on it's own line and give a bit of space to the methods

CalumMcCall:
I just gave this branch a smoke test and it worked, so I think it's ready to merge.

PR #287, opened 2017-12-18
[AN-5845] fix for availability messages being counted as unread

makingthematrix:
This is a quick fix for the unread dot: If an availability status is received by an old client, it is counted as a message of the type UNKNOWN. It does not appear anywhere, but is still counted as unread.

Here I just added one more check to counting unread messages - if the message is of the type UNKNOWN, it's not counted. (Status changes are already filtered out in the new clients, so they are not counted at all). We should also implement to clear all UNKNOWN messages already in the database.

APK for testing: https://drive.google.com/open?id=1rNBHVZ7Kn7BtWfmOSGc4-o-gVz3CkpQO


someoneEsle:
Bug 1
After upgrading to this apk, I still have the unread dot

makingthematrix:
This part of the fix only filters out unknown messages from conversations with other people, but to get rid of the unread dot we need also to filter out the self-conversation. That is done in the UI: https://github.com/wireapp/wire-android/pull/1337 .
Both PRs are necessary for the fix.

PR #292, opened 2018-01-16
[AN-5886] Add request for team invitation

JPonte:


tests/unit/src/test/scala/com/waz/service/InvitationServiceSpec.scala
```diff
@@ -0,0 +1,91 @@
+/*
+ * Wire
+ * Copyright (C) 2016 Wire Swiss GmbH
```
deanrobertcook: hmm...

zmessaging/src/main/scala/com/waz/sync/client/InvitationClient.scala
```diff
@@ -82,4 +98,18 @@ object InvitationClient {
...
+
+  case class ConfirmedTeamInvitation(id: InvitationId, emailAddress: EmailAddress, createdAt: Instant, teamId:TeamId)
+  object ConfirmedTeamInvitation {
+    def unapply(resp: ResponseContent): Option[ConfirmedTeamInvitation] = resp match {
+      case JsonObjectResponse(js) =>
+        Try(ConfirmedTeamInvitation(
+          InvitationId(js.getString("id")),
+          EmailAddress(js.getString("email")),
+          Instant.parse(js.getString("created_at")),
+          TeamId(js.getString("team"))
```
deanrobertcook: Maybe use the `JsonDecoder._` import?

PR #297, opened 2018-01-30
[AN-5752] Fix bounds checking in LzwDecoder

CalumMcCall:
This PR adds some missing bounds checking as pointed out in the security review.

Another approach to this would be changing the types from signed to unsigned, but this seemed simpler. @marcoconti83 can you take a look at this? My C++ is pretty rusty.

For the 3rd and final issue stated in the report, the review states that values other than 31 could cause overflow, but I couldn't replicate this in testing, so I think this is erroneous. `clear = 0` should never cause the check to pass. However, overflow/underflow is one of those super tricky things so I'd appreciate some other eyes on this.

### DIFF
--- a/zmessaging/src/main/jni/LzwDecoder.cpp
+++ b/zmessaging/src/main/jni/LzwDecoder.cpp
@@ -28,22 +28,32 @@
 #define LOG(...) printf(__VA_ARGS__)
 #endif
 
-static const int MAX_STACK_SIZE = 4096;
-static const int PIXEL_STACK_SIZE = 8192;
-static const int NULL_CODE = -1;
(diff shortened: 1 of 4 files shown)


zmessaging/src/main/jni/LzwDecoder.cpp
```diff
@@ -34,6 +34,9 @@ static const int NULL_CODE = -1;
 
 
 LzwDecoder::LzwDecoder(byte* image, int* pixels, int* colors, int width, int height) {
+	if (image == NULL || width < 0 || height < 0) {
```
marcoconti83: Would it make sense to use `<=` instead of `width < 0`? Does an image that is 0 pixel wide make sense?

PR #302, opened 2018-02-01
[AN-5920] allow for checking of current authentication cookie

deanrobertcook:


deanrobertcook:
Will discuss with raphael if this is actually better than previously before merging

PR #303, opened 2018-02-01
[AN-5921] Remove all-zero check in randombytes.c

deanrobertcook:


zmessaging/src/main/scala/com/waz/utils/crypto/RandomBytes.scala
```diff
@@ -34,7 +34,7 @@ class RandomBytes {
     val buffer = Array.ofDim[Byte](count)
 
     Try(randomBytes(buffer, count)).toOption match {
```
CalumMcCall: Does this `randomBytes()` call also cause the call to `JNI_OnLoad`? If not, then we can probably leave out this match block entirely.

PR #304, opened 2018-02-01
[AN-5751] Added canaries to ndk libs

CalumMcCall:
- Also remove spotify deps as we don't use them anymore


build.sbt
```diff
@@ -200,164 +168,6 @@ lazy val testutils = project.in(file("tests") / "utils")
...
-//      "org.robolectric" % "android-all" % RobolectricVersion,
-//      Deps.avs,
-//      "org.threeten" % "threetenbp" % "1.3",
-//      "com.wire.cryptobox" % "cryptobox-jni" % "0.8.2",
-//      "com.android.support" % "support-v4" % supportLibVersion % Provided,
-//      "com.google.android.gms" % "play-services-base" % "7.8.0" % Provided exclude("com.android.support", "support-v4"),
-//      "com.google.android.gms" % "play-services-gcm" % "7.8.0" % Provided
-//    )
-//  )
-
```
deanrobertcook: Please don't remove these uncommented parts of the build sbt - it's all there for the actors library on which QA depends, but it needs fixing - will be addressed in https://github.com/wireapp/wire-android-sync-engine/pull/290/files

CalumMcCall: Ah ok, I didn't realize. I've removed those changes.

PR #309, opened 2018-02-11
Move search logic down to UserSearchService

deanrobertcook:
Removes some unnecessary helper classes and brings all the search controller logic down to the search service. Still need to fix the tests

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
@@ -17,13 +17,15 @@
  */
 package com.waz.service
 
-import com.waz.ZLog._
 import com.waz.ZLog.ImplicitTag._
+import com.waz.ZLog._
 import com.waz.content.{MembersStorage, MessagesStorage, SearchQueryCacheStorage, UsersStorage}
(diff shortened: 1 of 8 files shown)


zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -77,25 +88,78 @@ class UserSearchService(selfUserId:           UserId,
...
+    if (filter.isEmpty) Future.successful {
       System.gc() // TODO: [AN-5497] the user search should not create so many objects to trigger GC in-between
     }
 
-    val topUsersSignal: Signal[IndexedSeq[UserData]] =
-      if (searchState.shouldShowTopUsers(teamId.isDefined))
-        topPeople.map(_.filter(u => !excludedUsers.contains(u.id) && !u.isWireBot)).map(SeqMap(_)(_.id, identity).values)
-      else Signal.const(IndexedSeq.empty[UserData])
+    def moveToFront(results: IndexedSeq[UserData]): IndexedSeq[UserData] = {
+      def toLower(str: String) = Locales.transliteration.transliterate(str).trim.toLowerCase
```
JPonte: The transliterator seems to be already lowering: "Any-Latin; Latin-ASCII; Lower; [^\ 0-9a-z] Remove"

PR #325, opened 2018-03-21
Fix asset retention

CalumMcCall:
This fixes our asset retention policies so they conform to the ones in the docs. This involved some renaming, which is done in the second commit for clarity, but should probably be merged into one once approved. Also some refactoring.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
+++ b/zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
@@ -90,8 +90,10 @@ object AssetClient {
 
   sealed abstract class Retention(val value: String)
   object Retention {
-    case object Eternal extends Retention("eternal")
+    case object Eternal extends Retention("eternal") //Only used for profile pics currently
+    case object EternalInfrequentAccess extends Retention("eternal-infrequent_access")
     case object Persistent extends Retention("persistent")
+    case object Expiring extends Retention("expiring")
     case object Volatile extends Retention("volatile")
   }
 
--- a/zmessaging/src/main/scala/com/waz/content/NotificationStorageImpl.scala
+++ b/zmessaging/src/main/scala/com/waz/content/NotificationStorageImpl.scala
@@ -26,7 +26,6 @@ import com.waz.utils.events.{AggregatingSignal, EventStream, Signal}
 import com.waz.utils.{CachedStorage, CachedStorageImpl, TrimmingLruCache}
 
 import scala.collection._
-import scala.concurrent.Future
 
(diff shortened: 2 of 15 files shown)


tiago-loureiro:
Just noticed this one; note that we have not pushed these changes yet so we should probably not merge this in before https://github.com/wireapp/wire-server/pull/252 gets merged.

CalumMcCall:
We still need to write the migration code to trigger the slow sync on the clients, so we aren't quite ready yet anyway.

zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
```diff
@@ -88,10 +91,28 @@ object AssetClient {
 
   val AssetsV3Path = "/assets/v3"
 
+  private def checkConv(convId: ConvId, members: MembersStorage, users: UserService) =
```
deanrobertcook: probably better to make this an inlined function inside the method `retentionPolicy`. The signature `checkConv: Signal[Boolean]` doesn't tell me much, and it's specific only to `retentionPolicy`

zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
```diff
@@ -88,10 +91,28 @@ object AssetClient {
...
+    members
+      .activeMembers(convId)
```
deanrobertcook: Also, since you're always calling head from this `Signal` you're returning, you could always just call `getActiveMembers` and this method could return a `Future` instead

zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
```diff
@@ -88,10 +91,28 @@ object AssetClient {
...
+      .flatMap(p => Signal.sequence(p.map(users.userSignal).toSeq:_*)
+        .map(_.exists(_.teamId.isDefined)))
+
+  def retentionPolicy(teamId: Option[TeamId], convData: ConversationData, members: MembersStorage, users: UserService): Signal[Retention] =
```
deanrobertcook: And finally, the companion object to `AssetClient` is probably not the best place to put this method. You could put it somewhere like the `AssetService` where you can inject `Option[TeamId]`, `MembersStorage` and `UserService`

zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
CalumMcCall: Ah good shout, that's much cleaner!

PR #332, opened 2018-04-09
Database import. Database export refactored.

yeryomenkom:


zmessaging/src/main/scala/com/waz/service/AccountManager.scala
```diff
@@ -54,20 +50,34 @@ import scala.util.Right
...
+    ))(userPrefs.remove)
+
   val storage   = global.factory.baseStorage(userId)
   val db        = storage.db
   val userPrefs = storage.userPrefs
 
   val account     = global.accountsStorage.signal(userId)
-  val clientState = userPrefs(SelfClient).signal
+  val clientState = for {
+    _ <- Signal.future(doAfterBackupCleanup())
```
JPonte: Shouldn't we check for `startedJustAfterBackup` before running the cleanup?

zmessaging/src/main/scala/com/waz/service/AccountsService.scala
```diff
@@ -215,23 +217,46 @@ class AccountsServiceImpl(val global: GlobalModule) extends AccountsService {
     _   <- Future.sequence(ids.map(enterAccount(_, None)))
   } yield Serialized.future(AccountManagersKey)(Future[Unit](accountManagers.mutateOrDefault(identity, Set.empty[AccountManager])))
 
-  override def enterAccount(userId: UserId, importDbFile: Option[File]) = {
-    //TODO import dbFile and create ZMessagingDb
-    Serialized.future(AccountManagersKey) {
+//  override def enterAccount(userId: UserId, importDbFile: Option[File]) = {
```
JPonte: Remove comments before merging.

zmessaging/src/main/scala/com/waz/service/AccountManager.scala
yeryomenkom: Sure.

PR #335, opened 2018-04-13
Add hasPassword request

CalumMcCall:
- Use this to display more accurate options


zmessaging/src/main/scala/com/waz/service/AccountManager.scala
```diff
@@ -269,6 +269,12 @@ class AccountManager(val userId:   UserId,
       case Left(err)                                  => CancellableFuture.successful(Left(err))
     }
   }
+
+  def hasPassword(): CancellableFuture[Option[Boolean]] =
```
deanrobertcook: You can just pass the `ErrorOrResponse[Boolean]` directly up to the UI - or if you want to tidy up a bit, just `ErrorOr[Boolean]`. The UI uses the error codes to decide what error message to display to the user and give them a bit of info about the problem

PR #340, opened 2018-04-18
Clarify errors in AuthenticationManager

CalumMcCall:


zmessaging/src/main/scala/com/waz/api/impl/ErrorResponse.scala
```diff
@@ -47,11 +47,13 @@ object ErrorResponse {
   val UnverifiedCode = 497
   val TimeoutCode = 599
   val ConnectionErrorCode = 598
+  val UnauthorizedCode = 401
 
   val InternalError = ErrorResponse(InternalErrorCode, "InternalError", "")
   val Cancelled = ErrorResponse(CancelledCode, "Cancelled", "")
   val Unverified = ErrorResponse(UnverifiedCode, "Unverified", "")
   val PasswordExists = ErrorResponse(Forbidden, "Forbidden", "password-exists")
+  val Unauthorized = ErrorResponse(UnauthorizedCode, "Unauthorized", "account-logged-out")
```
deanrobertcook: I don't know if all "Unauthorized" responses mean you're logged out, but since we want to tidy this up, let's just leave it for now

CalumMcCall: Ok good point.

PR #341, opened 2018-04-19
Batch load notifications from storage

CalumMcCall:


zmessaging/src/main/scala/com/waz/db/Dao.scala
```diff
@@ -152,6 +152,10 @@ abstract class BaseDao[T] extends Reader[T] {
     }
   }
 
+  def countWithDefaultSql(col: String, default: Long = -1): String = s"SELECT COALESCE(MAX(${col}), $default) FROM ${table.name}"
+
+  def maxWithDefault(sql: String)(implicit db: DB): Long = withStatement(sql)(_.simpleQueryForLong())
```
JPonte: The names of these methods seem misleading. Isn't `countWithDefaultSql ` the `maxWithDefault ` and `maxWithDefault ` just a query for a `Long`?

CalumMcCall: Aye good catch, these names are wack. I'll change it.

zmessaging/src/main/scala/com/waz/service/push/PushNotificationEventsStorage.scala
```diff
@@ -75,41 +75,39 @@ class PushNotificationEventsStorageImpl(context: Context, storage: Database, cli
...
-          pn.events
-            .toVector
-            .zipWithIndex
-            .map { case (obj, index) =>
-              PushNotificationEvent(pn.id, index, decrypted = false, obj, transient = pn.transient)
-            }
-        } else {
-          warn(s"Got notification ${pn.id} containing no events for us, ignoring")
-          Seq.empty[PushNotificationEvent]
+        pn.events.toVector.collect { case event =>
```
JPonte: What's the reason for using collect rather than map here, since your partial function matches everything?

PR #343, opened 2018-04-24
Migrate to OkHttp websockets.

yeryomenkom:


zmessaging/src/main/scala/com/waz/service/push/WSPushService.scala
```diff
@@ -0,0 +1,189 @@
...
+  def deactivate(): Unit
+  def notifications(): EventStream[Seq[PushNotificationEncoded]]
+  def connected(): Signal[Boolean]
+  def awaitActive(): Future[Unit]
+}
+
+object WSPushServiceImpl {
+
+  type AuthTokenProvider = () => ErrorOr[AccessToken]
+  type RequestCreator = AccessToken => okhttp3.Request
```
yeryomenkom: Abstract over all okhttp classes inside WSPushServiceImpl.

zmessaging/src/main/scala/com/waz/threading/CancellableFuture.scala
```diff
@@ -232,6 +232,54 @@ object CancellableFuture {
     }
   }
 
+  def timeout(duration: Duration,
```
yeryomenkom: Unused code. Could be useful in the future, but we can remove it.

PR #347, opened 2018-05-03
Update AVS API

CalumMcCall:


zmessaging/src/main/scala/com/waz/service/call/CallingService.scala
```diff
@@ -304,13 +307,27 @@ class CallingService(val accountId:       UserId,
...
       vbr <- userPrefs.preference(UserPreferences.VBREnabled).apply()
+      members <- members.getActiveUsers(conv.id)
     } yield {
+      val callType = if (isVideo && members.size > 4) {
+          Avs.WCallType.ForcedAudio
+        } else if (isVideo) {
+          Avs.WCallType.Video
+        } else {
+          Avs.WCallType.Normal
+        }
```
deanrobertcook: feel free to drop the braces

zmessaging/src/main/scala/com/waz/service/call/CallingService.scala
```diff
@@ -304,13 +307,27 @@ class CallingService(val accountId:       UserId,
       else if (conv.team.isEmpty) Future.successful(Set(UserId(conv.id.str)))
       else members.getByConvs(Set(conv.id)).map(_.map(_.userId).filter(_ != accountId).toSet)
       vbr <- userPrefs.preference(UserPreferences.VBREnabled).apply()
+      members <- members.getActiveUsers(conv.id)
```
deanrobertcook: notice 2 lines up we seem to want to the list of members as well. Maybe we can do this call above and also use the result in that final `else` block.

zmessaging/src/main/scala/com/waz/service/call/CallingService.scala
```diff
@@ -304,13 +307,27 @@ class CallingService(val accountId:       UserId,
...
+          Avs.WCallType.Video
+        } else {
+          Avs.WCallType.Normal
+        }
+
+      val convType = if (members.size == 2) {
+        Avs.WCallConvType.OneOnOne
+      } else {
+        Avs.WCallConvType.Group
+      }
```
deanrobertcook: I don't think this is correct - we probably want to use the same sense of what a `group` is as is used in the rest of the app. Even if a conversation has only 2 members, it can still be a "group" conversation, as other members can be added/removed. Notice above we already have the `isGroup` result which does a more comprehensive check

PR #362, opened 2018-06-12
Public logging

deanrobertcook:
With some other tidy-up work.

Still to do before this is safe:
* Ensuring all 'sub'classes of UUID are obfuscated
* lots of testing


zmessaging/src/main/scala/com/waz/model/Handle.scala
```diff
@@ -19,7 +19,73 @@ package com.waz.model
...
 import java.util.UUID
 
-import com.waz.utils.Locales
+import com.waz.log.InternalLog.ProductionLoggable
+import com.waz.utils.{Locales, sha2}
+
+import scala.language.implicitConversions
+import scala.math.Ordering
+
+case class SafeString(str: String) extends ProductionLoggable {
```
JPonte: Why are these in the `Handle.scala` file?

zmessaging/src/main/scala/com/waz/model/ConversationData.scala
```diff
@@ -58,11 +59,43 @@ case class ConversationData(id:                   ConvId              = ConvId()
                             accessRole:           Option[AccessRole]  = None,
                             link:                 Option[Link]        = None) {
```
JPonte: Maybe this link should also be a safe string?

deanrobertcook:
https://github.com/wireapp/android-project/issues/129

PR #367, opened 2018-06-27
Improve performance of creating large conversations and syncing users

makingthematrix:
The most important change here is simply increasing the limit for user ids sent in the "create conversation" request. It's now set to 256 which means that we practically won't use the "add members to the conversation" request at this point, as 256 is the conversation size limit anyway. 

Also, in merging sync jobs, I changed the `forall` check to `subsetOf` - it does the same but is optimized for hash sets, so we can expect a bit better performance.

The limit for the number of user ids in one sync request was increased from 45 to 64. It doesn't do much, but again, it might improve performance a little (fewer requests). Increasing this limit more than that results in a failure response (uri too large). 

~~I created methods `syncUsers` in `UserService`, and `take(number: Int)` in `StorageDao` and `CachedStorage` - they can be used in tests.~~

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/sync/SyncRequest.scala
+++ b/zmessaging/src/main/scala/com/waz/model/sync/SyncRequest.scala
@@ -17,8 +17,8 @@
  */
 package com.waz.model.sync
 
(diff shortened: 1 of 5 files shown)


makingthematrix:
build again

PR #371, opened 2018-07-02
Optimize user sync

makingthematrix:


zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -218,7 +195,6 @@ class UserServiceImpl(selfUserId:        UserId,
         sync.syncUsers(id)
         Signal.empty[UserData]
       case Some(data) =>
-        syncIfNeeded(data)
```
deanrobertcook: again, I think we can get rid of the `userSignal` from here and just use the storage directly

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -250,36 +226,41 @@ class UserServiceImpl(selfUserId:        UserId,
       case _ => syncSelfNow
     }
 
-  override def getUsers(ids: Seq[UserId]): Future[Seq[UserData]] =
-    usersStorage.listAll(ids) map { users =>
-      syncIfNeeded(users: _*)
-      users
-    }
+  override def getUsers(ids: Seq[UserId]): Future[Seq[UserData]] = usersStorage.listAll(ids)
```
deanrobertcook: user storage directly?

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -397,6 +378,8 @@ class UserServiceImpl(selfUserId:        UserId,
 object UserService {
   val DefaultUserName: String = ""
 
+  val SyncIfOlderThanSeconds = 300L
```
deanrobertcook: use scala.Duration

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -250,36 +226,41 @@ class UserServiceImpl(selfUserId:        UserId,
...
+  }
 
-  override def updateSyncedUsers(users: Seq[UserInfo], timestamp: Long = System.currentTimeMillis()): Future[Set[UserData]] = {
+  val currentConvMembers = for {
+    Some(convId) <- stats.selectedConversationId
+    membersIds   <- membersStorage.activeMembers(convId)
+    members      <- Signal.sequence(membersIds.map(usersStorage.signal).toSeq:_*)
+  } yield members
+
+  currentConvMembers(users => syncIfNeeded(users))
```
deanrobertcook: define and listen to signals in the top of the class

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -250,36 +226,41 @@ class UserServiceImpl(selfUserId:        UserId,
...
-    * Schedules user data sync if stored user timestamp is older than last slow sync timestamp.
+    * Schedules user data sync for guests and unconnected users if their timestamp is old
    */
-  override def syncIfNeeded(users: UserData*): Future[Option[SyncId]] =
-    lastSlowSyncTimestamp() flatMap {
-      //TODO: Remove empty picture check when not needed anymore
-      case Some(time) => sync.syncUsersIfNotEmpty(users.filter(user => user.syncTimestamp < time || user.picture.isEmpty).map(_.id))
-      case _ => sync.syncUsersIfNotEmpty(users.filter(_.picture.isEmpty).map(_.id))
-    }
+  override def syncIfNeeded(users: Seq[UserData], olderThan: Instant = Instant.now().minusSeconds(SyncIfOlderThanSeconds)): Future[Option[SyncId]] = {
```
deanrobertcook: should take a `Set[UserId]`

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -88,47 +90,30 @@ trait UserService {
...
-  * TODO improve accuracy of sync logic wrt to connected and unconnected users
-  * Currently, we sync all users on full-sync or if, when retrieving them via this class, we detect it's been a while since
-  * their last sync. This is both inefficient and incorrect. An improvement would be:
-  * 1. Since we get update events for all connected/team users, we don't need to bother syncing them outside of a full-sync.
-  * 2. For unconnected && non-team users, we should monitor the current conversation and just sync all unconnected users in
-  *    that conversation (there won't be that many on average, but maybe we'd need a throttle of a few minutes?). We should
-  *    also merge this with the ExpiredUsersService below, since that's what we do with wireless users, except we have a
-  *    timer for them.
-  * 3. Finally, we should listen to the self signals of all other logged in accounts and update our user's storage to reflect those states
+  * TODO: we should listen to the self signals of all other logged in accounts and update our user's storage to reflect those states
```
deanrobertcook: this should be relatively quick to do, you event have the `AccountsService` already injected

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -250,36 +226,41 @@ class UserServiceImpl(selfUserId:        UserId,
...
-    }
+  override def getUsers(ids: Seq[UserId]): Future[Seq[UserData]] = usersStorage.listAll(ids)
 
   /**
    * Schedules user data sync if user with given id doesn't exist or has old timestamp.
    */
   override def syncNotExistingOrExpired(users: Seq[UserId]): Future[Option[SyncId]] = usersStorage.listAll(users) flatMap { found =>
     val toSync = (users.toSet -- found.map(_.id)).toSeq
-    if (toSync.nonEmpty) sync.syncUsers(toSync: _*) flatMap (sId => syncIfNeeded(found: _*).map(_.orElse(Some(sId)))) else syncIfNeeded(found: _*)
+    if (toSync.nonEmpty) sync.syncUsers(toSync: _*) flatMap (sId => syncIfNeeded(found).map(_.orElse(Some(sId)))) else syncIfNeeded(found)
```
deanrobertcook: You could combine this method with the `syncIfNeeded`, which would then also need to check to see whether we have the `UserData` available locally, or else add the id to the list of ids to sync

deanrobertcook:
Also, the sync method `syncUsersIfNotEmpty(Set[UserId])` seems to have the exact same result as `syncUsers(Set[UserId])`. You could modify `syncUsers` slightly and then just use that everywhere

tests/unit/src/test/scala/com/waz/log/InternalLogSpec.scala
```diff
@@ -331,6 +331,7 @@ class InternalLogSpec extends AndroidFreeSpec {
       InternalLog.reset()
 
       val paths = pathsInTempDir
+      println(paths)
```
deanrobertcook: what is this?

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -88,47 +85,50 @@ trait UserService {
...
+  val currentConvMembers = for {
+    Some(convId) <- stats.selectedConversationId
+    membersIds   <- membersStorage.activeMembers(convId)
+    members      <- Signal.sequence(membersIds.map(usersStorage.signal).toSeq:_*)
+  } yield members
+
+  currentConvMembers(users => syncIfNeeded(users.map(_.id).toSet))
+
+  //Update user data for other accounts
+  //TODO remove this and move the necessary user data up to the account storage
```
deanrobertcook: how can we move the user data up to the account storage?

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -250,36 +216,26 @@ class UserServiceImpl(selfUserId:        UserId,
...
-  /**
-    * Schedules user data sync if stored user timestamp is older than last slow sync timestamp.
-   */
-  override def syncIfNeeded(users: UserData*): Future[Option[SyncId]] =
-    lastSlowSyncTimestamp() flatMap {
-      //TODO: Remove empty picture check when not needed anymore
-      case Some(time) => sync.syncUsersIfNotEmpty(users.filter(user => user.syncTimestamp < time || user.picture.isEmpty).map(_.id))
-      case _ => sync.syncUsersIfNotEmpty(users.filter(_.picture.isEmpty).map(_.id))
+  */
+  override def syncIfNeeded(userIds: Set[UserId], olderThan: Instant = Instant.now().minusSeconds(SyncIfOlderThanSeconds.toSeconds)): Future[Option[SyncId]] =
```
deanrobertcook: I think `olderThan` should be a `FiniteDuration`, and the calculation should be moved into the body. 

Plus, don't use `Instant.now()`, but rather `clock.instant()` (improves testability) and we have `RichInstant` and `RichFiniteDuration` which you can import for time calculations:
`clock.instant() - olderThan`

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -250,36 +216,26 @@ class UserServiceImpl(selfUserId:        UserId,
...
+    usersStorage.listAll(userIds).flatMap { found =>
+      val newIds = userIds -- found.map(_.id)
+      val existing = found.filter(u => !u.isConnected && (u.teamId.isEmpty || u.teamId != teamId) && u.syncTimestamp.forall(_.isBefore(olderThan)))
+      val toSync = newIds ++ existing.map(_.id)
+      verbose(s"syncIfNeeded for users; new: (${newIds.size}) + existing: (${existing.size}) = all: (${toSync.size})")
+      if (toSync.nonEmpty) sync.syncUsers(toSync).map(Some(_))(Threading.Background) else Future.successful(None)
     }
 
-  override def updateSyncedUsers(users: Seq[UserInfo], timestamp: Long = System.currentTimeMillis()): Future[Set[UserData]] = {
+  override def updateSyncedUsers(users: Seq[UserInfo], timestamp: Instant = Instant.now()): Future[Set[UserData]] = {
```
deanrobertcook: maybe call this parameter `syncTime`

zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -88,47 +85,50 @@ trait UserService {
...
-  }
+    }
+
+  val currentConvMembers = for {
+    Some(convId) <- stats.selectedConversationId
+    membersIds   <- membersStorage.activeMembers(convId)
+    members      <- Signal.sequence(membersIds.map(usersStorage.signal).toSeq:_*)
+  } yield members
+
+  currentConvMembers(users => syncIfNeeded(users.map(_.id).toSet))
```
deanrobertcook: seems a bit pointless to get all the users to then map back to their ids. `currentConvMembers` could just be a `Signal[Set[UserId]]` and then if in any place we need the full data, we add the last step there.

zmessaging/src/main/scala/com/waz/service/UserService.scala
makingthematrix: This is an old TODO I copied together with the code. I don't think it applies anymore. I'll remove it.

PR #380, opened 2018-07-10
Fix ErrorResponse model.

yeryomenkom:


zmessaging/src/main/scala/com/waz/api/impl/ErrorResponse.scala
```diff
@@ -65,19 +65,15 @@ object ErrorResponse {
...
-  //TODO Add proper construction function
   implicit val errorResponseConstructor: CustomErrorConstructor[ErrorResponse] =
     new CustomErrorConstructor[ErrorResponse] {
       override def constructFrom(error: HttpClient.HttpClientError): ErrorResponse = error match {
         case HttpClient.EncodingError(err) =>
           ErrorResponse.InternalError.copy(message = s"Encoding error: $err")
         case HttpClient.DecodingError(err, response) =>
-          val bodyStr = BodyDeserializer[String].deserialize(response.body)
-          ErrorResponse.InternalError.copy(message = s"Decoding body error: $err \nBody string: $bodyStr")
+          ErrorResponse(response.code, label = "Decoding error", message = s"Decoding body error: $err")
```
JPonte: I don't think the label should be "Decoding error" since the response might have one regardless of an invalid body.

yeryomenkom: And what I have to put there? What is your's variant?

PR #395, opened 2018-08-03
Fix asset messages having their timestamp resent on resending

makingthematrix:
fixes https://github.com/wireapp/android-project/issues/176

### DIFF
--- a/zmessaging/src/main/scala/com/waz/sync/handler/AssetSyncHandler.scala
+++ b/zmessaging/src/main/scala/com/waz/sync/handler/AssetSyncHandler.scala
@@ -21,52 +21,46 @@ import com.waz.ZLog.ImplicitTag._
 import com.waz.ZLog._
 import com.waz.api.impl.ErrorResponse._
 import com.waz.cache.CacheService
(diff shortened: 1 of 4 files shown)


zmessaging/src/main/scala/com/waz/sync/handler/AssetSyncHandler.scala
```diff
@@ -37,36 +37,37 @@ class AssetSyncHandler(cache:   CacheService,
...
-        CancellableFuture successful Right(None)
+        CancellableFuture successful Left(internalError("No asset data found, got local data"))
       case (None, None) =>
         debug(s"No asset data found, no local data found")
-        CancellableFuture successful Right(None)
-      case _ =>
-        debug(s"Unexpected error")
-        CancellableFuture successful Right(None)
+        CancellableFuture successful Left(internalError("No asset data found, no local data found"))
+      case err =>
```
deanrobertcook: This isn't an err, but of type (Option[AssetData], Option[LocalData]). I think it's probably fine if you just collapse all these invalid states into one case block and just report some internal error, we'll be visiting this again shortly anyway

PR #410, opened 2018-08-17
Fix connection requests not opening and not appearing in the conv list

makingthematrix:


makingthematrix:
tbh, I'm not happy with that adding the other user to `memberStorage` triggers the conversation's sync job which finally results in the conversation being displayed correctly. It's very counterintuitive. But it's just a one line fix, so it will have to suffice.

zmessaging/src/main/scala/com/waz/service/ConnectionService.scala
```diff
@@ -117,6 +117,7 @@ class ConnectionServiceImpl(selfUserId:      UserId,
     for {
       otoConvs <- convs.getOneToOneConversations(selfUserId, oneToOneConvData.toSeq)
       convToUser = eventMap.flatMap(e => otoConvs.get(e._1).map(c => c.id -> e._1))
+      _ <- Future.sequence(convToUser.map { case (convId, userId) => members.add(convId, Set(selfUserId, userId)) })
```
JPonte: It would be better to avoid `Future.sequence`. You could create a method `addAll` that inserts a bunch of members at once.

zmessaging/src/main/scala/com/waz/content/MembersStorage.scala
```diff
@@ -96,24 +97,35 @@ class MembersStorageImpl(context: Context, storage: ZmsDatabase) extends CachedS
...
       case (convId, users) => users.map(ConversationMemberData(_, convId))
     }
 
     removeAll(removeList).zip(insertAll(addList)).map(_ => ())
   }
 
+  def addAll(members: Map[ConvId, Set[UserId]]): Future[Unit] = getActiveUsers2(members.keySet).flatMap { active =>
+    val addList =
+      members
+        .transform((convId, users) => users -- active.getOrElse(convId, Set()))
```
JPonte: Not sure this optimisation is worth another db query (`getActiveUsers2`).

PR #418, opened 2018-09-03
Force create add member message for unexpected users

deanrobertcook:


zmessaging/src/main/scala/com/waz/service/messages/MessagesContentUpdater.scala
```diff
@@ -113,21 +114,11 @@ class MessagesContentUpdater(messagesStorage: MessagesStorage,
...
-            else { // msg was already synced, need to create new local message
-              shouldCreate = true
-              msg
-            }
-          } flatMap { res =>
-            verbose(s"shouldCreate: $shouldCreate")
-            if (shouldCreate) addLocalMessage(create).map(Some(_))
-            else Future.successful(res)
-          }
+          verbose(s"got local message: $msg, is local?: ${msg.isLocal}")
```
JPonte: Revert this changes as discussed

PR #421, opened 2018-09-07
Fix duplicated entries in Conversations due to the connect request bug

makingthematrix:


zmessaging/src/main/scala/com/waz/service/conversation/ConversationsContentUpdater.scala
```diff
@@ -83,6 +84,16 @@ class ConversationsContentUpdaterImpl(val storage:     ConversationStorage,
...
 
+  private val shouldFixDuplicatedConversations = userPrefs.preference(UserPreferences.FixDuplicatedConversations)
+
+  for {
+    shouldFix <- shouldFixDuplicatedConversations()
+  } if (shouldFix) {
+    fixDuplicatedConversations().flatMap(_ =>
+      shouldFixDuplicatedConversations := false
+    )
+  }
```
deanrobertcook: perhaps:

```
for {
  shouldFix <- shouldFixDuplicatedConversations()
  _ <- if (shouldFix) fixDuplicatedConversations() else Future.successful({})
  _ <- if (shouldFix) shouldFixDuplicatedConversations := false else Future.successful({})
} yield {}
```

(you could even drop the second if check here)

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsContentUpdater.scala
```diff
@@ -213,4 +224,30 @@ class ConversationsContentUpdaterImpl(val storage:     ConversationStorage,
...
-}
\ No newline at end of file
+
+  def fixDuplicatedConversations(): Future[Unit] =
+    for {
+      duplicates <- storage.duplicatedConversations
+      convIds    = duplicates.flatten
+      users      <- usersStorage.getAll(convIds.map(id => UserId(id.str)))
+      isOriginal = convIds.zip(users.map(_.nonEmpty)).toMap
+      pairs      = duplicates.flatMap { pair =>                             // there should be exactly one original and one or more duplicates
```
deanrobertcook: I feel as though this comment should be above the `if upd.nonEmpty => Some(orig, upd)`, no?

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsContentUpdater.scala
```diff
@@ -213,4 +224,30 @@ class ConversationsContentUpdaterImpl(val storage:     ConversationStorage,
 
   override def updateAccessMode(id: ConvId, access: Set[Access], accessRole: Option[AccessRole], link: Option[ConversationData.Link] = None) =
     storage.update(id, conv => conv.copy(access = access, accessRole = accessRole, link = if (!access.contains(Access.CODE)) None else link.orElse(conv.link)))
-}
\ No newline at end of file
+
+  def fixDuplicatedConversations(): Future[Unit] =
+    for {
+      duplicates <- storage.duplicatedConversations
+      convIds    = duplicates.flatten
```
deanrobertcook: you may want to convert this to a `Set` here - by definition you'll have at least 2 of each (which I don't think you care about in `convIds`)

zmessaging/src/main/scala/com/waz/content/ConversationStorage.scala
```diff
@@ -116,6 +119,15 @@ class ConversationStorageImpl(storage: ZmsDatabase) extends CachedStorageImpl[Co
     updateSearchKey(cs collect { case (p, c) if p.name != c.name || (p.convType == Group) != (c.convType == Group) || (c.name.nonEmpty && c.searchKey.isEmpty) => c })
   }
 
+  def duplicatedConversations: Future[Iterable[Iterable[ConvId]]] = init.map { _ =>
+    conversationsById.values.groupBy(_.remoteId).filter(_._2.size > 1).map(_._2.map(_.id))
+  }
```
deanrobertcook: I feel as though you don't need this method - it's only used in one place and as a temporary piece of code - just call `ConversationStorage#list()` and do it in the `for` comprehension

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsContentUpdater.scala
```diff
@@ -213,4 +224,30 @@ class ConversationsContentUpdaterImpl(val storage:     ConversationStorage,
...
+        (original.headOption, updates) match {
+          case (Some(orig), upd) if upd.nonEmpty => Some(orig, upd)
+          case _ => None
+        }
+      }.toMap
+      _ = verbose(s"fixDuplicatedConversations: (original, updates) pairs: $pairs")
+      _ <- Future.sequence(pairs.map { case (origId, updates) =>
+            for {
+              _ <- if (updates.tail.nonEmpty) storage.removeAll(updates.tail) else Future.successful({})
+              _ <- storage.updateLocalId(origId, updates.head)
```
deanrobertcook: I think these are back-to-front, as discussed

deanrobertcook:
Some suggestions for your commits:
* Keep the titles a little shorter and to the point, the first one could be: "fix: merge duplicate conversations", and the second: "fix: avoid inconsistencies in remote conversation id mapping"
* In the description, talk about ways this bug was triggered, why you did the fix
* At the end of the commit, link to the tickets that were fixed

PR #423, opened 2018-09-11
Add users search for a specific conversation

JPonte:
This allows to perform search using a text query for participants in a conversation which will be needed for mentions.


JPonte:
As reported by @someoneEsle, this doesn't select conversation participants who aren't connected to the user.

JPonte:
build again

PR #425, opened 2018-09-14
feat: new Mentions

makingthematrix:
We already had `mentions` in the protobuf and handled it in `MessageContent`, but they were not used in the app. Now we want to introduce them properly, and that means changes both to how they are encoded in the protobuf, and new code which actually uses them.

New `Mention`s can be now sent from `ConversationsUiService.addTextMessage`.
They are added to `MessageContent`s of `MessageData` and adjusted for the specified charset.
We differentiate between two charsets: UTF-16 and default. When we want to send a message,
we specify that the mentions in that message should be adjusted for UTF-16. When we receive
one, we need to re-adjust the mentions for our default charset (UTF-8 on Android). Sending
and receiving messages use common fuctionality in MessageData: `messageContent`. Before this
method was used the same way no matter the direction. Now we have to specify which way
the message will go after adjusting. This creates a special case for editing messages.
An edited message is both re-displayed on the our side, as if we received it, and it is sent
to other devices. In this commit we handle only the sending side. The re-displaying of
an edited message needs additional work.

### DIFF
--- a/project/SharedSettings.scala
+++ b/project/SharedSettings.scala
@@ -26,7 +26,7 @@ object SharedSettings {
     lazy val avs = "com.wire" % "avs" % avsVersion
     lazy val avsAudio = "com.wire.avs" % "audio-notifications" % audioVersion
     lazy val cryptobox = "com.wire" % "cryptobox-android" % cryptoboxVersion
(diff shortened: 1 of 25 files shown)


makingthematrix:
Build again.

makingthematrix:
build again

PR #430, opened 2018-09-28
feat: Pass message mentions in the NotificationInfo

JPonte:
It will allow UI to highlight them in the notifications

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/Mention.scala
+++ b/zmessaging/src/main/scala/com/waz/model/Mention.scala
@@ -17,4 +17,23 @@
  */
 package com.waz.model
 
+import com.waz.utils.{JsonDecoder, JsonEncoder}
+import org.json.JSONObject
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/model/Mention.scala
```diff
@@ -17,4 +17,23 @@
  */
 package com.waz.model
 
+import com.waz.utils.{JsonDecoder, JsonEncoder}
+import org.json.JSONObject
+
 case class Mention(userId: Option[UserId], start: Int, length: Int)
+
+object Mention{
+  implicit lazy val Decoder: JsonDecoder[Mention] = new JsonDecoder[Mention] {
```
makingthematrix: With the decoder and encoder here, you can now remove methods `decodeMentions` and `encodeMentions` from `MessageData`.

PR #431, opened 2018-10-01
[WIP] refactor: Remove remoteMap from ConversationStorage

makingthematrix:
`remoteMap` is a mutable hashmap of remoteId -> localId pairs, working as a cache in `ConversationStorage`. Its purpose is to improve performance when looking for a conversation by remoteId. In order to do it, `remoteMap` has to be updated every time we add, remove, or update localId of any conversation. Recently it led to a serious bug when accepting new connection requests. The bug is fixed in another PR, but a mutable map in this crucial place, accessed and updated from many threads, is a constant threat. In this PR I try to remove it completely, together with the other cache, `conversationsById`, and replace it with directly accessing the database. The database is in-memory, so the performance shouldn't be worse.


deanrobertcook:
I'm not sure if you need an explicit guarantee that all conversations are always in memory, but the `ConversationStorage` class uses an `UnlimitedLRUCache`, and so even without the signal, every hit to the database should be put into the in-memory cache. So, if I'm not mistaken, you can actually just remove all this extra logic in the Storage class. The only reason we needed to do this before is because the `remoteMap` required such an explicit guarantee (which it seems, we didn't even properly uphold), and now we don't have it, we can test what happens if we just use the default Storage logic

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
```diff
@@ -356,9 +356,6 @@ class ConversationsUiServiceImpl(selfUserId:      UserId,
     } yield (conv, syncId)
   }
 
-  override def findGroupConversations(prefix: SearchKey, limit: Int, handleOnly: Boolean): Future[Seq[ConversationData]] =
-    convStorage.search(prefix, selfUserId, handleOnly).map(_.sortBy(_.displayName)(currentLocaleOrdering).take(limit))
```
deanrobertcook: Good man!

makingthematrix:
build again

makingthematrix:
build again

zmessaging/src/main/scala/com/waz/content/ConversationStorage.scala
```diff
@@ -90,78 +71,51 @@ class ConversationStorageImpl(storage: ZmsDatabase) extends CachedStorageImpl[Co
...
     verbose(s"${cs.size} convs deleted")
-    cs foreach { c =>
-      conversationsById.remove(c) foreach { cd =>
-        if (remoteMap(cd.remoteId) == c) {
-          convDeleted ! cd
-          remoteMap.remove(cd.remoteId)
-        }
-      }
+    contents.head.map(_.filterKeys(csSet.contains)).foreach { convs =>
+      convs.values.foreach(convDeleted ! _)
```
deanrobertcook: So.. this looks problematic. `onDeleted` is also used to update `contents`, so at the point we reach this line of code, will `contents` still contain the old deleted conversations? Seems like a race condition. We may need to think around this...

zmessaging/src/main/scala/com/waz/content/ConversationStorage.scala
```diff
@@ -26,61 +26,42 @@ import com.waz.model.ConversationData.ConversationType.Group
...
-  private val remoteMap         = new mutable.HashMap[RConvId, ConvId]()
-  private val conversationsById = new mutable.HashMap[ConvId, ConversationData]()
-
-  override val convsSignal      = Signal[ConversationsSet]()
-
-  def conversations             = conversationsById.values.toIndexedSeq
-
   val convAdded = EventStream[ConversationData]()
   val convDeleted = EventStream[ConversationData]()
   val convUpdated = EventStream[(ConversationData, ConversationData)]() // (prev, updated)
```
deanrobertcook: `convAdded` and `convUpdated` don't seem to be necessary anymore. 

`convDeleted` on the other hand will be a little trickier...

PR #432, opened 2018-10-02
feat: Change muted value from 2 to 3 states

makingthematrix:
https://github.com/wearezeta/documentation/pull/64/files

Now a conversation might be unmuted, muted, or it might by "mentions-only". Also, we want to have a possibility to easily expand this functionality in the future, so the new `muted` field is implemented as a set of flags. On the backend side it is a bitmask - an integer where each bit corresponds to one of the flags. For some time we will have to maintain the old `muted` boolean in requests and responses from the backend. The old `muted == true` corresponds to a `MuteMask` set with both `StandardMuted` and `MentionsMuted`.


JPonte:
build again

makingthematrix:
build again

PR #442, opened 2018-10-22
fix: use fileUri for broadcast to media scanner

CalumMcCall:
In upgrading to android 8, we moved all file uris (`file:///`) to
content uris (`content:///`) to be complient with changes introduced in
android 7, however these uris don't seem to work with the media scanner
and so our saved images stopped appearing in the system as images.

This creates 2 uris, one file type, to be used by the scanner, and
another content type, to be returned to other parts of the app that will
then share this URI with other applications via other types of intents.

Testing
  - Save an image on Android version >= 7, open the image from the
    generated notification and make sure it doesn't crash
  - Save an image on any android version, and make sure that image
    appears in the gallery (or any other application)


deanrobertcook:
build again

PR #443, opened 2018-10-22
fix: onIncomingCall should replace finished calls from the profile

JPonte:
onIncomingCall should update the call profile unless it's for a group check, which occurs on an ongoing call. Terminating or Ended calls should be discarded.


JPonte:
build again

PR #444, opened 2018-10-23
feat: Quotes handling in protobuf, storage, and the API for UI.

makingthematrix:
A `Quote` is defined in the protobuf as a field in the `Text` message, consisting of the `MessageId` of the original message, and a sha256 used to verify the quote. This PR is focused only on the messageId part.

There are two new fields in `MessageData`: `quote` and `quoteValidity`, and two new columns in the Messages table in the database. `quote` holds the message id of the original message, `quoteValidity` is a boolean confirming that the sha256 comparison was successful (default to false, not used yet).
There are also new methods in `ConversationsUiService` and `MessagesService` which can be used to reply to a message. They take the original message's id instead of the conversation's id, and first check if the original message exists, and then use the conversation's id of that message.

This change affects UI, because in `WebLinkPartView` we access links directly from MessageData's protos, and that part was changed. I added a new method to `MessageData`, `links`, which can now be used in `WebLinkPartView` instead.


makingthematrix:
For making the UI work with this PR, please go to `WebLinkPartView` and change
```
case TextMessage(_, _, previews) if index >= 0 && previews.size > linkIndex => Some(previews(linkIndex))
```
to
```
case TextMessage(_, _, previews, _) if index >= 0 && previews.size > linkIndex => Some(previews(linkIndex))

File: zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala [hunk 1]
```diff
@@ -155,20 +159,37 @@ class MessagesStorageImpl(context: Context,
...
       msgs.acquire { msgs =>
-        val unread = msgs.filter { m => !m.isLocal && m.convId == conv && m.time.isAfter(lastReadTime) && !m.isDeleted && m.userId != userId && m.msgType != Message.Type.UNKNOWN } .toVector
-        UnreadCount(
-          unread.count(m => !m.isSystemMessage && m.msgType != Message.Type.KNOCK && !m.hasMentionOf(userId)),
-          unread.count(_.msgType == Message.Type.MISSED_CALL),
-          unread.count(_.msgType == Message.Type.KNOCK),
-          unread.count(_.hasMentionOf(userId))
-        )
+        val unread = msgs.filter { m => !m.isLocal && m.convId == conv && m.time.isAfter(lastReadTime) && !m.isDeleted && m.userId != userId && m.msgType != Message.Type.UNKNOWN }.toVector
+        Future.sequence(unread.map(isQuoteButNotMention)).map(_.count(q => q)).map { unreadQuotes =>
```
deanrobertcook: @makingthematrix 
This could potentially be a very expensive line if there are lot of unread messages in a conversation. 

Try to think of a way to reduce the number of DB requests and if you get stuck I'll give a hint

zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
```diff
@@ -155,20 +159,37 @@ class MessagesStorageImpl(context: Context,
...
+          verbose(s"unread: ${unread.size}, unread quotes: $unreadQuotes")
+          UnreadCount(
+            unread.count(m => !m.isSystemMessage && m.msgType != Message.Type.KNOCK && !m.hasMentionOf(userId)) - unreadQuotes,
+            unread.count(_.msgType == Message.Type.MISSED_CALL),
+            unread.count(_.msgType == Message.Type.KNOCK),
+            unread.count(_.hasMentionOf(userId)),
+            unreadQuotes
```
deanrobertcook: perhaps name the parameters here for readability's sake

zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
```diff
@@ -155,20 +159,37 @@ class MessagesStorageImpl(context: Context,
...
+          )
+        }
       }
     }
+
   }
 
+  override def isQuoteOfSelf(msg: MessageData) =
+    msg.quote.fold(Future.successful(false))(quoteId => get(quoteId).map(_.exists(_.userId == userId)))
```
deanrobertcook: When batching is done, this method should be unnecessary

zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -210,7 +210,30 @@ class NotificationService(context:         Context,
...
+      )
+    }
+
+    val notifications = msgs.map(m =>
+      messages.isQuoteOfSelf(m).map(isQuote =>
+        build(m, isQuote, drift = pushService.beDrift.currentValue.getOrElse(Duration.Zero))
+      )
+    )
+
+    Future.sequence(notifications).map(_.flatten).foreach(add)
```
deanrobertcook: Batch here as well

zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -210,7 +210,30 @@ class NotificationService(context:         Context,
...
+        if (msg.time == RemoteInstant.Epoch) msg.localTime.toRemote(drift) else msg.time,
+        ephemeral = msg.isEphemeral,
+        mentions  = msg.mentions.flatMap(_.userId),
+        isQuote   = msg.quote.nonEmpty
+      )
+    }
+
+    val notifications = msgs.map(m =>
+      messages.isQuoteOfSelf(m).map(isQuote =>
+        build(m, isQuote, drift = pushService.beDrift.currentValue.getOrElse(Duration.Zero))
```
deanrobertcook: might be nicer to inline the `build` method

zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -210,7 +210,30 @@ class NotificationService(context:         Context,
     val updatedAssets = updates collect {
       case (prev, msg) if msg.state == Message.Status.SENT && msg.msgType == Message.Type.ANY_ASSET => msg
     }
-    if (updatedAssets.nonEmpty) add(updatedAssets.flatMap(m => notification(m, pushService.beDrift.currentValue.getOrElse(Duration.Zero))))
```
deanrobertcook: does removing the `nonEmpty` check perhaps trigger any unwanted updates to notifications? In the worst case this could lead to notifications "ticking" again. It's probably okay, but please be sure

makingthematrix: No unwanted updates. It's easier to see now, after the batching is in place.

makingthematrix:
build again

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
```diff
@@ -153,10 +166,15 @@ class ConversationsUiServiceImpl(selfUserId:      UserId,
     verbose(s"postAssetMessage: $convId, $asset")
     for {
       message    <- messages.addAssetMessage(convId, asset, exp)
+      _ = verbose(s"postAssetMessage 1")
```
JPonte: I think you forgot to remove these.

PR #445, opened 2018-10-26
feat: generate and check reply hashes

CalumMcCall:
- Remove dependency on android's base64 encoding/decoding and
   instead use a jvm, Java 7 compatible library


CalumMcCall:
build again

CalumMcCall:
build again

PR #447, opened 2018-11-06
feature: public logging. skeleton

yeryomenkom:
Compiler checked public logging without macros implemented with custom string interpolation and LogShow type class. Also added logLevel property to LogOutput trait.


deanrobertcook:
build again

PR #450, opened 2018-11-19
[WIP] refactor: Make remote time in MessageData an Option of RemoteInstant

makingthematrix:
The old val `time` is changed to a private val `remoteTime`. `time` becomes a method, so in most places in the code, where we only read the time, no changes are necessary. The `time` method checks `remoteTime` and if it's `None`, it returns `localTime` modified by the current BackEnd drift.


deanrobertcook:
closing as discussed

PR #455, opened 2018-11-28
fix: subscribe to global event context on signal head

deanrobertcook:
For cases where the underlying signal has no reference held elsewhere in
the app, the signal my have been garbage collected before the first value
is ever published to the listener that completes the promise. This would
mean the listener is never called, and the future would never complete.

In this PR, we instead create a regular subscription and subscribe to the
global event context, which will then keep a reference to the signal until
the promise is completed. We now also do not check the current value before
subscribing, since if there is a value, we get called back immediately
anyway. This also has the added bonus that calling head won't call
disableAutoWiring on the Signal, meaning it won't be alive indefinitely
(as long as a reference to it is held).


deanrobertcook:
didn't fix the problem - let's not fix what ain't broke

PR #457, opened 2018-12-11
Event-based notifications

deanrobertcook:


zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -17,374 +17,265 @@
...
+        _ <- storage.insertAll(toShow)
+        _ <- pushNotificationsToUi()
+      } yield {}
+    } else Future.successful({})
+  })
+
+  def pushNotificationsToUi(): Future[Unit] = {
+    verbose(l"pushNotificationsToUi")
+    for {
+      toShow         <- storage.list().map(_.toSet)
```
makingthematrix: No need to make it into a set. It's used as a seq everywhere.

PR #460, opened 2018-12-20
feat: Allow log in with an SSO id

makingthematrix:
There is a new field in the AccountData (and the ActiveAccounts table in db) for holding SSO ids. If the user is logged in via SSO, it overrides email, phone, and password - wherever these three are requested, if SSO Id is present, they are not needed anymore.


makingthematrix:
build again

PR #462, opened 2019-01-10
fix: Improvements for slowdown

JPonte:
The introduction of the work manager brought some shortcomings that we didn't expect, like no throttling or order guarantee of commands (which led to no proper way of merging them as well). 
This PR only removes the delay of the last read command and a predicate fix for read receipt messages. It is possible that other issues we haven't noticed yet will arise.


JPonte:
build again

PR #463, opened 2019-01-11
feature: added safe base64 object for android and jvm.

yeryomenkom:
Move base64 to private, static var. This facilitates running of tests which use base64(ReplyHashing) whilst also not requiring us to pass implicit bas64 params everywhere in the code base(lots of implicit json decoder objects require base64 for example). Still do to is to remove a hack Calum added to allow the above mentioned unit tests for ReplyHashing to run. Once that is done we can merge. Currently this isn't a priority so this PR will remain open until Calum has time.


zmessaging/src/main/scala/com/waz/utils/crypto/AESUtils.scala
```diff
@@ -36,8 +35,8 @@ object AESUtils {
   lazy val randomBytes = new RandomBytes()
 
   //TODO Remove usage of android Base64 class
```
JPonte: No longer TODO I guess?

CalumMcCall:
This is good to merge now

PR #464, opened 2019-01-14
feat: Enable customization of branding

CalumMcCall:
- This commit changes the backendConfig/firebase/cert pinning to
   facilitate softcoding of these parameters for clients, instead of
   hardcoding them


zmessaging/src/main/scala/com/waz/znet/ServerTrust.scala
```diff
@@ -47,6 +48,9 @@ object ServerTrust {
     0x7A, 0xBF, 0xE7, 0xF6, 0x27, 0x51, 0x0C, 0xD3, 0xE1, 0x63, 0xF6, 0xFB, 0x1E, 0x23, 0x64, 0xD8, 0xAD,
     0x02, 0x03, 0x01, 0x00, 0x01).map(_.toByte)
 
+  //TODO: This is just here to ensure SE can compile on its own. The actual pin is specified in default.json
```
JPonte: Is this an actual To Do or just a warning?

CalumMcCall: good catch!

PR #468, opened 2019-01-21
External partner role

yeryomenkom:


zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
```diff
@@ -106,20 +111,16 @@ object TeamsClient {
...
-        case JsonObjectResponse(js) if js.has("members") =>
-          Try(decodeSet('members)(js, TeamMemberDecoder)).toOption.map(_.toMap)
-        case _ =>
-          warn(s"Unexpected response: $response")
-          None
-      }
-    }
+  case class TeamMember(user: UserId, permissions: Option[Permissions])
+
+  case class Permissions(self: Long, copy: Long)
```
CalumMcCall: Can you explain what `self` and `copy` represent? I don't understand how this represents a bitmask

zmessaging/src/test/scala/com/waz/sync/client/TeamsClientSpec.scala
```diff
@@ -20,8 +20,11 @@ package com.waz.sync.client
 import com.waz.model.AccountDataOld._
 import com.waz.model.AccountDataOld.Permission._
 import com.waz.specs.AndroidFreeSpec
+import com.waz.sync.client.TeamsClient.TeamMembers
+import com.waz.utils.CirceJSONSupport
 
-class TeamsClientSpec extends AndroidFreeSpec {
+//TODO Replace with integration test when AuthRequestInterceptor2 will be used
```
CalumMcCall: "... when AuthRequestInterceptor2 is introduced"

zmessaging/src/test/scala/com/waz/sync/client/PushTokenClientSpec.scala
```diff
@@ -21,9 +21,7 @@ import com.waz.ZIntegrationSpec
 
 class PushTokenClientSpec extends ZIntegrationSpec {
 
-  import com.waz.AuthenticationConfig._
-
-  //TODO reintroduce test when AuthRequestInterceptor2 is used
+  //TODO reintroduce test when AuthRequestInterceptor2 will be used
```
CalumMcCall: see other comment below

zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
yeryomenkom: It is legacy names. From my understanding 'self' is your permissions mask and 'copy' is permissions that you can setup/change for other team members.

PR #477, opened 2019-01-28
Permissions improvements

yeryomenkom:
Move all permissions and role storing logic into new service. Use that service everywhere where we need to check/store permissions/roles.

Maybe it worth to create our own set of permissions. It looks like backend permissions do not fit well our ui logic.


marcoconti83:
@makingthematrix @johnxnguyen should we merge this, or not relevant anymore?

marcoconti83:
build again

johnxnguyen:
@marcoconti83 I think it's still relevant, but it looks like it still needs to be hooked up.

PR #480, opened 2019-01-30
fix: Send a typing indicator event

makingthematrix:
Recent introduction of `RequestInfo` broke the typing indicator.

In `TypingSyncHandler` we check if the request is not too much delayed, by checking the job's `startTime`. But due to other requirements, if there is no delay at all, we have the `startTime` set to 0, not to `now()`, which meant that the request was always delayed too much. I added a check for that.


makingthematrix:
build again

makingthematrix:
build again

PR #483, opened 2019-02-06
fix: messages which were deleted still showing up

CalumMcCall:
- When opening a cleared conversation from the searchUI
   messages from before the clearing would still show.
 - This commit adds a missing notification of changes which resulted in the cached `MessagesListView` displaying deleted content
 - Fixes wireapp/android-project#391


CalumMcCall:
build again

PR #485, opened 2019-02-07
Fix video downloads

JPonte:


zmessaging/src/main/scala/com/waz/service/assets2/DownloadAssetStorage.scala
```diff
@@ -54,7 +54,7 @@ object DownloadAssetStorage {
     val UploadStatus   = asInt(_.status)('status)
 
     override val idCol = Id
-    override val table = Table("DownloadAssets", Id, Mime, Preview, Size, Details)
+    override val table = Table("DownloadAssets", Id, Mime, Name, Preview, Details, Downloaded, Size, UploadStatus)
```
yeryomenkom: Maybe DownloadStatus instead of UploadStatus?

zmessaging/src/main/scala/com/waz/service/assets2/models.scala
```diff
@@ -162,17 +162,38 @@ case class DownloadAsset(
 
 object DownloadAsset {
 
-  def create(asset: GenericAsset): DownloadAsset =
+  def create(asset: GenericAsset): DownloadAsset = {
```
yeryomenkom: why input parameter so general? 
for example: I can not imagine situation when we can construct DownloadAsset from UploadAsset.

zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
```diff
@@ -76,7 +76,28 @@ class MessageEventProcessor(selfUserId:          UserId,
     } yield standard ++ updatedQuotes
   }
 
+  case class EventAndLocalData(event: MessageEvent, message: Option[MessageData], asset: Option[DownloadAsset])
+
+  def localDataForEvent(event: MessageEvent): Future[EventAndLocalData] = {
+    event match {
+      case GenericMessageEvent(_, _, _, c) => storage.get(MessageId(c.messageId)).flatMap {
+        case Some(message) => message.assetId match {
```
yeryomenkom: Can we rewrite it without pattern matching?

zmessaging/src/main/scala/com/waz/service/assets2/DownloadAssetStorage.scala
JPonte: It's tricky actually. Because it's the upload status of the sender of the asset we will download. But I think it's better to call it DownloadAssetStatus.

zmessaging/src/main/scala/com/waz/service/assets2/models.scala
JPonte: This is not GeneralAsset, it's the proto.

zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
JPonte: Sure

zmessaging/src/main/scala/com/waz/service/assets2/models.scala
yeryomenkom: true. sorry

zmessaging/src/main/scala/com/waz/service/assets2/DownloadAssetStorage.scala
yeryomenkom: So... Do you agree about 'DownloadStatus' name?

JPonte: `AssetStatus` is better. It belongs to `DownloadAsset` so we already know it's the status from the `DownloadAsset`.

yeryomenkom: Or even Status. I agree.

PR #489, opened 2019-02-13
feat: add parsing of managedBy flag from BE

CalumMcCall:
- This flag indicates whether the user should be allowed to edit their
   profile or not, and so we have to store it locally.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
+++ b/zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
@@ -28,14 +28,14 @@ object ManagedBy {
     case u => Unknown(u)
   }
 
-  sealed trait ManagedBy extends Product with Serializable
-  case object Wire extends ManagedBy {
+  sealed trait ManagedBy
+  final case object Wire extends ManagedBy {
     override def toString: String = "wire"
   }
-  case object SCIM extends ManagedBy {
+  final case object SCIM extends ManagedBy {
     override def toString: String = "scim"
   }
-  sealed case class Unknown(str: String) extends ManagedBy {
+  final case class Unknown(str: String) extends ManagedBy {
     override def toString: String = str
   }
 
--- a/zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
+++ b/zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
@@ -1,0 +1,46 @@
+/*
+ * Wire
+ * Copyright (C) 2016 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.model
+
+import com.waz.utils.JsonDecoder.decodeOptString
+import org.json.JSONObject
+
+object ManagedBy {
+
+  def apply(s: String): ManagedBy = s match {
+    case "wire" => Wire
+    case "scim" => SCIM
+    case u => Unknown(u)
+  }
+
+  case object Wire extends ManagedBy {
+    override def toString: String = "wire"
+  }
+  case object SCIM extends ManagedBy {
+    override def toString: String = "scim"
+  }
+  sealed case class Unknown(str: String) extends ManagedBy {
+    override def toString: String = str
+  }
+
+  def decodeOptManagedBy(s: Symbol)(implicit js: JSONObject): Option[ManagedBy] = decodeOptString(s).map(ManagedBy(_))
+}
+
+sealed trait ManagedBy extends Product with Serializable
+
+
--- a/zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
+++ b/zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
@@ -28,6 +28,7 @@ object ManagedBy {
     case u => Unknown(u)
   }
 
+  sealed trait ManagedBy extends Product with Serializable
   case object Wire extends ManagedBy {
     override def toString: String = "wire"
   }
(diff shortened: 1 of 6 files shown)


CalumMcCall:
build again

zmessaging/src/main/scala/com/waz/model/SSOId.scala
```diff
@@ -41,7 +40,7 @@ object SSOId {
     override def apply(implicit js: JSONObject): SSOId = decodeSSOId(js)
   }
 
-  implicit lazy val Encoder: JsonEncoder[SSOId] = new JsonEncoder[SSOId] {
+  implicit object Encoder extends JsonEncoder[SSOId] {
```
marcoconti83: Should this be part of this PR? Sounds related to SSO ID instead

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -234,16 +239,17 @@ object UserData {
...
     override val table = Table(
       "Users", Id, TeamId, Name, Email, Phone, TrackingId, Picture, Accent, SKey, Conn, ConnTime, ConnMessage,
-      Conversation, Rel, Timestamp, DisplayName, Verified, Deleted, AvailabilityStatus, Handle, ProviderId, IntegrationId, ExpiresAt
+      Conversation, Rel, Timestamp, DisplayName, Verified, Deleted, AvailabilityStatus, Handle, ProviderId, IntegrationId, ExpiresAt, ManagedBy
     )
 
     override def apply(implicit cursor: DBCursor): UserData = new UserData(
       Id, TeamId, Name, Email, Phone, TrackingId, Picture, Accent, SKey, Conn, RemoteInstant.ofEpochMilli(ConnTime.getTime), ConnMessage,
-      Conversation, Rel, Timestamp, DisplayName, Verified, Deleted, AvailabilityStatus, Handle, ProviderId, IntegrationId, ExpiresAt
+      Conversation, Rel, Timestamp, DisplayName, Verified, Deleted, AvailabilityStatus, Handle, ProviderId, IntegrationId, ExpiresAt, ManagedBy.map(model.ManagedBy(_))
```
marcoconti83: Why is `ManagedBy` the only filed that requires a `map(...)` in this list? Other fields such as email or handle don't have it

zmessaging/src/main/scala/com/waz/model/SSOId.scala
CalumMcCall: I decided to change it to match the decoder definition. I can remove it though if you like.

zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
```diff
@@ -0,0 +1,46 @@
...
+    override def toString: String = "scim"
+  }
+  sealed case class Unknown(str: String) extends ManagedBy {
+    override def toString: String = str
+  }
+
+  def decodeOptManagedBy(s: Symbol)(implicit js: JSONObject): Option[ManagedBy] = decodeOptString(s).map(ManagedBy(_))
+}
+
+sealed trait ManagedBy extends Product with Serializable
```
makingthematrix: `Product` and `Serializable` are probably not necessary. 
And, for better readability, please put the trait `ManagedBy` directly above case objects with extend it.

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -52,14 +52,16 @@ case class UserData(override val id:       UserId,
...
+                    expiresAt:             Option[RemoteInstant] = None,
+                    managedBy:             Option[ManagedBy]     = None) extends Identifiable[UserId] {
 
   def isConnected = ConnectionStatus.isConnected(connection)
   def hasEmailOrPhone = email.isDefined || phone.isDefined
   def isSelf = connection == ConnectionStatus.Self
   def isAcceptedOrPending = connection == ConnectionStatus.Accepted || connection == ConnectionStatus.PendingFromOther || connection == ConnectionStatus.PendingFromUser
   def isVerified = verified == Verification.VERIFIED
   def isAutoConnect = isConnected && ! isSelf && connectionMessage.isEmpty
+  def isReadOnlyProfile = managedBy.nonEmpty && !managedBy.contains(ManagedBy.Wire)
```
makingthematrix: You can use `managedBy.exists(_ != ManagedBy.Wire)`.

zmessaging/src/main/scala/com/waz/model/UserData.scala
CalumMcCall: Yeah you're right, I've simplified it and I'm just testing now

zmessaging/src/main/scala/com/waz/model/ManagedBy.scala
CalumMcCall: Ah, yeah you're right, moving ManagedBy into the companion object also fixed a linting issue around names being hidden. I think the `Product` thing is good practice though, it helps prevent weird type inference, as explained [here](https://nrinaudo.github.io/scala-best-practices/adts/product_with_serializable.html) but maybe I'm applying this wrong.

zmessaging/src/main/scala/com/waz/model/UserData.scala
CalumMcCall: Unfortunately, I can't, because if the `managedBy` is None, we want `isReadOnlyProfile` to be false. Both `contains` and `exists` return `false` if the option is empty, which isn't what we want.

zmessaging/src/main/scala/com/waz/model/SSOId.scala
marcoconti83: From a technical point of view might make total sense, others can chime in on this aspect. I'm just pointing out that the PR is about ManagedBy, so I would keep anything that is not about ManagedBy out, in case this change has side effects that we don't expect.

CalumMcCall: Yeah, I was just thinking the same thing. This is clearly an area where I need to be more disciplined! :)

marcoconti83: Thanks for looking into this!

marcoconti83: More thoughts on the topic: https://medium.com/@fagnerbrack/one-pull-request-one-concern-e84a27dfe9f1

zmessaging/src/main/scala/com/waz/model/UserData.scala
makingthematrix: `exists` in this case will return `true` if and only if the option is non-empty, but its value is different from `Wire`.

marcoconti83: @makingthematrix is right. Try this: https://scalafiddle.io/sf/D8GEGRB/0

marcoconti83:
build again

PR #490, opened 2019-02-15
feat: Add column to UserData for profile fields

CalumMcCall:
This can't be easily tested, due to this not being finished on the backend side, but for now there is a sync job(`SyncRichInfo` which takes a list of user ids) for fetching RichInfo. I'll be working on adding tests to try and make sure the JSON parsing works at least.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/UserData.scala
+++ b/zmessaging/src/main/scala/com/waz/model/UserData.scala
@@ -88,7 +88,7 @@ case class UserData(override val id:       UserId,
     expiresAt = user.expiresAt.orElse(expiresAt),
     teamId = user.teamId.orElse(teamId),
     managedBy = user.managedBy.orElse(managedBy),
-    fields = if(user.fields.isEmpty) fields else user.fields.get
+    fields = user.fields.getOrElse(fields)
   )
 
   def updated(user: UserSearchEntry): UserData = copy(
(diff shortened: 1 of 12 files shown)


zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -32,29 +33,30 @@ import org.json.JSONObject
...
+                    displayName:           Name                   = Name.Empty,
+                    verified:              Verification           = Verification.UNKNOWN, // user is verified if he has any otr client, and all his clients are verified
+                    deleted:               Boolean                = false,
+                    availability:          Availability           = Availability.None,
+                    handle:                Option[Handle]         = None,
+                    providerId:            Option[ProviderId]     = None,
+                    integrationId:         Option[IntegrationId]  = None,
+                    expiresAt:             Option[RemoteInstant]  = None,
+                    managedBy:             Option[ManagedBy]      = None,
+                    fields:                Option[Seq[UserField]] = None) extends Identifiable[UserId] {
```
JPonte: Is there any difference between a `None` and an empty `Seq` here?

CalumMcCall: Great point, yeah I think this can just be a `Seq`

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -88,7 +88,7 @@ case class UserData(override val id:       UserId,
     expiresAt = user.expiresAt,
     teamId = user.teamId.orElse(teamId),
     managedBy = user.managedBy.orElse(managedBy),
-    fields = user.fields.orElse(fields)
+    fields = if(user.fields.isEmpty) fields else user.fields.get
```
JPonte: `user.fields.getOrElse(fields)`?

marcoconti83:
@CalumMcCall any testing that we can do at this level?

marcoconti83:
I would not do backend testing at this stage, but I think it would be great to have json serialization unit testing. I added some tests already with https://github.com/wireapp/wire-android-sync-engine/pull/494 and we could just expand on those.

CalumMcCall:
build again

PR #492, opened 2019-02-19
Archive a conversation when the user leaves remotely

marcoconti83:
When the user leaves the conversation from another device, the current device should archive the conversation. This is not happening at the moment.

# Changes
When the app receives a `conversation.member-leave` that removes the currently logged in user, and the user was the one that triggered the event, then the conversation is archived.

# Testing
- Unit tests for a positive case (should archive) and a negative cases:
  - should not archive because triggered by a different user
  - should not archive because it's not the user that is being removed
- Manual test on device:
  - Leave the conversation from another device: conversation archived ✅
  - Have another user remove the user from another device: conversation not archived ✅

# Left for the future
The constructor for `ConversationServiceSpec` has too many parameters and depends on so many services. Mocking them is not easy. Can we come up with a better design?


marcoconti83:
build again

marcoconti83:
build again

PR #494, opened 2019-02-20
Fix `expires_at` not being preserved when updating `UserData`

marcoconti83:
Bug: wireless guests are not displayed as wireless. They have no expiration time and it's possible to connect to them.

The reason is that `expires_at`, which is used to distinguish if a user is a guest user, is overwritten with `None` when updating the `UserData`.

# Changes
- Fix `expires_at` field in `UserData` being overwritten with `None` when applying `UserInfo` to `UserData`
- Removed code duplication in the `apply` method of `UserData`
- Added tests to cover `UserData` update from UserInfo
- The new test showed that `teamId` was also not correctly initialized when creating `UserData` from `UserInfo`. This was fixed too by removing code duplication

# Testing
- Added new unit tests for `UserData` and `UserInfo`
- Manually tested on device before and after: now wireless users are displayed correctly

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/UserData.scala
+++ b/zmessaging/src/main/scala/com/waz/model/UserData.scala
@@ -83,7 +83,7 @@ case class UserData(override val id:       UserId,
     },
     providerId = user.service.map(_.provider),
     integrationId = user.service.map(_.id),
-    expiresAt = user.expiresAt,
+    expiresAt = user.expiresAt.orElse(expiresAt),
     teamId = user.teamId.orElse(teamId),
     managedBy = user.managedBy.orElse(managedBy)
   )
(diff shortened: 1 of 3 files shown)


zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -83,7 +83,7 @@ case class UserData(override val id:       UserId,
     },
     providerId = user.service.map(_.provider),
```
JPonte: Weren't these also wrong? Or do you prefer to fix that in a separate PR?

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -168,10 +168,7 @@ object UserData {
 
   def apply(user: UserInfo): UserData = apply(user, withSearchKey = true)
 
-  def apply(user: UserInfo, withSearchKey: Boolean): UserData =
-    UserData(user.id, None, user.name.getOrElse(Name.Empty), user.email, user.phone, user.trackingId, user.mediumPicture.map(_.id),
-      user.accentId.getOrElse(AccentColor.defaultColor.id), SearchKey(if (withSearchKey) user.name.getOrElse(Name.Empty).str else ""), deleted = user.deleted,
-      handle = user.handle, providerId = user.service.map(_.provider), integrationId = user.service.map(_.id))
+  def apply(user: UserInfo, withSearchKey: Boolean): UserData = UserData(user.id, user.name.getOrElse(Name.Empty)).updated(user, withSearchKey)
```
JPonte: `withSearchKey: Boolean = true` and you can delete the line above.

marcoconti83: It's not allowed apparently. This is what I get from the compiler:
```
[error] UserData.scala:130: in object UserData, multiple overloaded alternatives of method apply define default arguments.
```

zmessaging/src/main/scala/com/waz/model/UserData.scala
marcoconti83: I forgot about those. I will fix this in this PR.

marcoconti83:
build again

marcoconti83:
build again

PR #497, opened 2019-02-21
stylish refactoring in public logging.

yeryomenkom:


zmessaging/src/main/scala/com/waz/log/ZLog2.scala
```diff
@@ -342,6 +319,8 @@ object ZLog2 {
         """.stripMargin
       }
 
+    LogShow[Map[UserId, Option[LocalInstant]]]
+
```
johnxnguyen: What is the purpose of this?

yeryomenkom: Should be removed. Thanks

PR #499, opened 2019-02-26
Fix: edited messages are now marked as sent

marcoconti83:
Previously, the ID of the message would change during an edit. The code was looking for a message with a given ID to mark it as `SENT`, but the ID was already changed and no message was found, leaving the message in the `PENDING` state.

This only affected the sender, the receiver would receive the edit.

# Changes
- Propagate the ID of the message that has been sent. In case of edits, the message ID is changed and we are now propagating the new ID.

# Testing
I could not easily test this at the unit test level. The amount of mock necessary to test this is quite demanding. I manually tested on device before and after the fix. Steps:
- Open a conversation
- Enter a text message
- Send and wait until it's marked as "sent"
- Edit the message
- Send

Result before the fix:
- The edited message is shown as "pending" even after a minute

Result after the fix:
- The edited message is shown as "sent" within a couple of seconds

# Left for the future
We should work on making tests less dependent on complicated mock setup.


CalumMcCall:
build again

CalumMcCall:
build again

PR #501, opened 2019-03-01
Refactor: json user search client

johnxnguyen:
# What's In This PR?
We want to remove our own custom JSON (de)serialization and instead rely on **Circe**. This is a small step toward that goal.

The deserialization code in `UserSearchClient` has been replaced with auto derived circe json objects.  We define a response object for each backend response that we expect. In this case `UserSearchResponse` and `ExtractHandleResponse`. These objects reflect the structure of the response payload (but we only include keys that we're interested in). Simply by defining these objects, deserialization handled for us.

The model objects extracted from these responses are defined in the domain layer, to maintain separation. In this case, `UserSearchEntry` has been moved from `UserSearchClient` to `UserSearchService`.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
@@ -29,8 +29,8 @@ import com.waz.service.ZMessaging.clock
 import com.waz.service.conversation.{ConversationsService, ConversationsUiService}
 import com.waz.service.teams.TeamsService
 import com.waz.sync.SyncServiceHandle
-import com.waz.sync.client.UserSearchClient.UserSearchEntry
-import com.waz.threading.Threading
+import com.waz.sync.client.UserSearchClient.UserSearchResponse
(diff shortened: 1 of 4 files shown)


zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -331,4 +333,23 @@ object UserSearchService {
...
+    def apply(searchUser: UserSearchResponse.User): UserSearchEntry = {
+      import searchUser._
+      UserSearchEntry(UserId(id), Name(name), accent_id, Handle(handle))
+    }
+  }
+
+  /**
+    * Extracts `UserSearchEntry` objects contained within the given search response.
+    */
+  def unpack(response: UserSearchResponse): Seq[UserSearchEntry] = {
```
makingthematrix: Please rename it to `unapply`. This will be more idiomatic in Scala and let us use some syntax sugar in `match/case`, returning values from functions, etc.

https://docs.scala-lang.org/tour/extractor-objects.html

PR #504, opened 2019-03-06
feat: Users other than `self` should have the permission field in the user data

makingthematrix:
## What's new in this PR?

The functionality for filtering out from the user search results which should not be visible for partners. The results visible for standard team members are also slightly changed, taking into account that some of the results are partners.

For details please refer to the documentation:
https://github.com/wearezeta/documentation/blob/master/topics/teams/use%20cases/clients/101-search-for-partners.md

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/Handle.scala
+++ b/zmessaging/src/main/scala/com/waz/model/Handle.scala
@@ -25,11 +25,11 @@ case class Handle(string: String) extends AnyVal {
   override def toString : String = string
 
   def startsWithQuery(query: String): Boolean = {
-     string.startsWith(Handle.stripSymbol(query).toLowerCase)
+    query.nonEmpty && string.startsWith(Handle.stripSymbol(query).toLowerCase)
   }
 
   def exactMatchQuery(query: String): Boolean = {
-    string == Handle.stripSymbol(query).toLowerCase
+    query.nonEmpty && string == Handle.stripSymbol(query).toLowerCase
   }
 
   def withSymbol: String = if (string.startsWith("@")) string else s"@$string"
--- a/zmessaging/src/main/scala/com/waz/model/Handle.scala
+++ b/zmessaging/src/main/scala/com/waz/model/Handle.scala
@@ -24,13 +24,13 @@ import com.waz.utils.Locales
 case class Handle(string: String) extends AnyVal {
   override def toString : String = string
 
-  def startsWithQuery(query: String): Boolean = {
+  def startsWithQuery(query: String): Boolean =
     query.nonEmpty && string.startsWith(Handle.stripSymbol(query).toLowerCase)
-  }
+  
 
-  def exactMatchQuery(query: String): Boolean = {
-    query.nonEmpty && string == Handle.stripSymbol(query).toLowerCase
-  }
+  def exactMatchQuery(query: String): Boolean =
+    string == Handle.stripSymbol(query).toLowerCase
+
 
   def withSymbol: String = if (string.startsWith("@")) string else s"@$string"
 }
--- a/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
@@ -17,11 +17,14 @@
  */
 package com.waz.service
 
+import com.waz.ZLog
 import com.waz.ZLog.ImplicitTag._
+import com.waz.content.UserPreferences.SelfPermissions
 import com.waz.content._
 import com.waz.log.ZLog2._
 import com.waz.model.SearchQuery.{Recommended, RecommendedHandle}
 import com.waz.model.UserData.{ConnectionStatus, UserDataDao}
+import com.waz.model.UserPermissions.{PartnerPermissions, decodeBitmask}
 import com.waz.model.{SearchQuery, _}
 import com.waz.service.ZMessaging.clock
 import com.waz.service.conversation.{ConversationsService, ConversationsUiService}
@@ -42,6 +45,13 @@ case class SearchResults(top:   IndexedSeq[UserData]         = IndexedSeq.empty,
                          convs: IndexedSeq[ConversationData] = IndexedSeq.empty,
                          dir:   IndexedSeq[UserData]         = IndexedSeq.empty) { //directory (backend search)
   override def toString = s"SearchResults(top: ${top.size}, local: ${local.size}, convs: ${convs.size}, dir: ${dir.size})"
+
+  def isEmpty: Boolean = top.isEmpty && local.isEmpty && convs.isEmpty && dir.isEmpty
+}
+
+object SearchResults {
+  implicit val SearchResultsLogShow: LogShow[SearchResults] =
+    LogShow.create(sr => s"SearchResults(${sr.top.length}, ${sr.local.length}, ${sr.convs.length}, ${sr.dir.length})")
 }
 
 class UserSearchService(selfUserId:           UserId,
@@ -56,7 +66,9 @@ class UserSearchService(selfUserId:           UserId,
                         messages:             MessagesStorage,
                         convsStorage:         ConversationStorage,
                         convsUi:              ConversationsUiService,
-                        conversationsService: ConversationsService) {
+                        conversationsService: ConversationsService,
+                        userPrefs:            UserPreferences
+                       ) {
 
   import Threading.Implicits.Background
   import com.waz.service.UserSearchService._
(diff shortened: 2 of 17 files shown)


zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -67,34 +79,48 @@ class UserSearchService(selfUserId:           UserId,
...
-    searchLocal(filter).map(_.filter(u => !(u.isGuest(teamId) && teamOnly)))
+  private lazy val isPartner = userPrefs(SelfPermissions).apply()
+    .map(decodeBitmask)
+    .map(ps => PartnerPermissions.subsetOf(ps) && PartnerPermissions.size == ps.size)
+
+  private def filterForPartner(query: Filter, searchResults: IndexedSeq[UserData]): Future[IndexedSeq[UserData]] = {
+    lazy val knownUsers = membersStorage.getByUsers(searchResults.map(_.id).toSet).map(_.map(_.userId).toSet)
+    isPartner.flatMap {
+      case true if teamId.isDefined =>
+        ZLog.verbose(s"filterForPartner1 Q: $query, RES: ${searchResults.map(_.getDisplayName)}) with partner = true and teamId")
```
marcoconti83: How does this interact with the new logging system? The display name should not go into public logs

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -67,34 +79,48 @@ class UserSearchService(selfUserId:           UserId,
...
+        knownUsers.map(knownUsersIds => searchResults.filter(u => knownUsersIds.contains(u.id)))
+      case false if teamId.isDefined =>
+        ZLog.verbose(s"filterForPartner2 Q: $query, RES: ${searchResults.map(_.getDisplayName)}) with partner = false and teamId")
```
marcoconti83: Log

zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -64,6 +64,15 @@ class TeamsServiceImpl(selfUser:           UserId,
 
   private implicit val dispatcher = SerialDispatchQueue()
 
+  private val shouldSyncTeam = userPrefs.preference(UserPreferences.ShouldSyncTeam)
+
+  for {
+    shouldSync <- shouldSyncTeam()
+  } if (shouldSync && teamId.isDefined) {
+    verbose(l"Syncing the team $teamId")
```
marcoconti83: Log

zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
```diff
@@ -113,7 +119,10 @@ object TeamsClient {
 
   case class TeamMembers(members: Seq[TeamMember])
 
-  case class TeamMember(user: UserId, permissions: Option[Permissions])
+  case class TeamMember(user: UserId, permissions: Option[Permissions], created_by: Option[UserId])
+
+  implicit val TeamMemberLogShow: LogShow[TeamMember] =
+    LogShow.create(tm => s"TeamMember(${tm.user}, ${tm.permissions}, ${tm.created_by})")
```
marcoconti83: Log leaking user ID?

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
marcoconti83: From 1:1 discussion: it won't, `ZLog.verbose` only works if compiling in debug

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
marcoconti83: See above for resolution

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -67,34 +79,48 @@ class UserSearchService(selfUserId:           UserId,
   private val exactMatchUser = new SourceSignal[Option[UserData]]()
   private val signalMap = mutable.HashMap[SearchQuery, Signal[IndexedSeq[UserData]]]()
 
-  def usersForNewConversation(filter: Filter = "", teamOnly: Boolean): Signal[IndexedSeq[UserData]] =
-    searchLocal(filter).map(_.filter(u => !(u.isGuest(teamId) && teamOnly)))
+  private lazy val isPartner = userPrefs(SelfPermissions).apply()
+    .map(decodeBitmask)
+    .map(ps => PartnerPermissions.subsetOf(ps) && PartnerPermissions.size == ps.size)
```
CalumMcCall: see comment above about this maybe being simplified to simple calling `.equals()`

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -67,34 +79,48 @@ class UserSearchService(selfUserId:           UserId,
...
+        }
+      case _ => Future.successful(searchResults)
+    }
+  }
+
+  def usersForNewConversation(filter: Filter = "", teamOnly: Boolean): Signal[IndexedSeq[UserData]] = {
+    val isHandle = Handle.isHandle(filter)
+    searchLocal(filter, isHandle = isHandle)
+      .map(_.filter(u => !(u.isGuest(teamId) && teamOnly)))
+      .flatMap(res => Signal.future(filterForPartner(filter, res)))
```
CalumMcCall: My understanding is we are creating Signals from data which is unlikely to ever change. It feels like `usersForNewConversation` and `usersToAddToConversation` might be better off returning `Future` instead of `Signal`. Then we can get rid of these ugly `Signal.future` wraps, doing away with the subscriptions to them also. What do you think?

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -225,27 +247,35 @@ class UserSearchService(selfUserId:           UserId,
...
-      local <- searchLocal(filter, showBlockedUsers = true)
-      convs <- conversations
-      dir   <- directorySearch
+      top       <- topUsers
+      isPartner <- Signal.future(isPartner)
+      local     <- searchLocal(filter, showBlockedUsers = true, isHandle = isHandle)
+      local     <- Signal.future(filterForPartner(filter, local))
+      convs     <- conversations
+      dir       <- if (isPartner) Signal.const(IndexedSeq.empty[UserData]) else directorySearch
+      dir       <- Signal.future(filterForPartner(filter, dir))
```
CalumMcCall: I think it might be nicer to merge these last two lines. I'm surprised you can reuse variable names in a for comprehension. I think a small method encapsulating these two lines and flatmapping them together would be more readable, otherwise maybe just using different variable names for both these lines? As it stands it seems a bit weird. Also again, maybe this whole method should return a `Future` if we are simply wrapping `filterForPartner` again?

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -225,27 +247,35 @@ class UserSearchService(selfUserId:           UserId,
...
 
     for {
-      updated <- userService.updateUsers(results)
-      _       <- userService.syncIfNeeded(updated.map(_.id), Duration.Zero)
-      ids     = results.map(_.id)(breakOut): Vector[UserId]
-      _       = verbose(l"updateSearchResults($query, ${results.map(_.handle)})")
-      _       <- queryCache.updateOrCreate(query, updating(ids), SearchQueryCache(query, clock.instant(), Some(ids)))
+      userData <- usersStorage.listAll(ids)
+      filtered <- filterForPartner(query.filter, userData)
+      filteredIds = filtered.map(_.id).toSet
```
CalumMcCall: These two lines can be combined, as we never use `filtered` elsewhere

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -255,21 +285,20 @@ class UserSearchService(selfUserId:           UserId,
...
         queryCache.updateOrCreate(query, updating(userId), SearchQueryCache(query, clock.instant(), Some(Vector(userId))))
-    }(Threading.Background)
+    }
 
     Future.successful({})
   }
 
   def searchUserData(query: SearchQuery): Signal[IndexedSeq[UserData]] = signalMap.getOrElseUpdate(query, returning( startNewSearch(query) ) { _ =>
-    CancellableFuture.delay(cacheExpiryTime).map { _ =>
+    CancellableFuture.delay(cacheRefreshInterval).map { _ =>
```
CalumMcCall: What's the reason behind this change? Looks like we are "expiring" instead of "refreshing" here.

zmessaging/src/main/scala/com/waz/sync/handler/UsersSyncHandler.scala
```diff
@@ -40,7 +40,8 @@ class UsersSyncHandler(assetSync: AssetSyncHandler,
                        assets: AssetService,
                        usersClient: UsersClient,
                        imageGenerator: ImageAssetGenerator,
-                       otrSync: OtrSyncHandler) {
+                       otrSync: OtrSyncHandler
+                      ) {
```
CalumMcCall: This change doesn't seem necessary, can we roll it back?

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
makingthematrix: Yes, but that will propagate changes deep into UI which right now uses signals. After working on this I agree futures would be a better solution. After all, if I search for something, I don't really expect the results of my search to change while I'm looking at it. If I want to refresh it, I can do by simply searching again.
Unfortunately, we don't have time to rewrite it all right now. I started a separate PR for that, but it will take more time than we have.

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
CalumMcCall: We could simply rename the second instance of `dir` to `dirResults` or something. My real issue is just this reuse of variable names.

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
makingthematrix: Removed, together with the `signalMap`.

zmessaging/src/main/scala/com/waz/model/Handle.scala
```diff
@@ -25,11 +25,11 @@ case class Handle(string: String) extends AnyVal {
   override def toString : String = string
 
   def startsWithQuery(query: String): Boolean = {
-     string.startsWith(Handle.stripSymbol(query).toLowerCase)
+    query.nonEmpty && string.startsWith(Handle.stripSymbol(query).toLowerCase)
   }
 
   def exactMatchQuery(query: String): Boolean = {
-    string == Handle.stripSymbol(query).toLowerCase
+    query.nonEmpty && string == Handle.stripSymbol(query).toLowerCase
```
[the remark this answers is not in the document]
makingthematrix: `.startsWith("")` returns `true` and that's why `.nonEmpty` is needed in `startsWithQuery`, but it's not needed in `exactMatchQuery`. Fixed.

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -297,14 +296,8 @@ class UserSearchService(selfUserId:           UserId,
...
-  def searchUserData(query: SearchQuery): Signal[IndexedSeq[UserData]] = signalMap.getOrElseUpdate(query, returning( startNewSearch(query) ) { _ =>
-    CancellableFuture.delay(cacheRefreshInterval).map { _ =>
-      signalMap.remove(query)
-      queryCache.remove(query)
-    }
-  })
-
-  private def startNewSearch(query: SearchQuery): Signal[IndexedSeq[UserData]] = returning( queryCache.optSignal(query) ){ _ =>
+  // not private for tests
+  def searchUserData(query: SearchQuery): Signal[IndexedSeq[UserData]] = returning( queryCache.optSignal(query) ){ _ =>
```
CalumMcCall: Was this caching code you removed not needed anymore?

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -17,7 +17,7 @@
  */
 package com.waz.service
 
-import com.waz.ZLog
+import com.waz.ZLog.verbose
```
CalumMcCall: Should we maybe just migrate to ZLog2? You can use `showString` to display normal strings, to save you having to create instances of LogShow

zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
makingthematrix: Some time ago we suspected that because the user search sometimes returns a lot of results, it may trigger GC  and freeze the app. One way to avoid it was to call `System.gc()` before even querying. The other was to cache the signals instead of creating new ones - that's why the `signalMap` with the expiry time for signals.
But as I test it now it seems that it does not help. I would like to remove `signalMap` (and this call to `gc`) and see what happens in Internal and Beta. If we don't get any feedback about slower search (I guess we don't) it means it's not needed anymore.
In the meantime we created a lot of batch versions for methods in services and handlers. Instead of calling `add*` or `update*` many times for each instance of a case class, we call a batch method for the whole sequence of them. It might be the reason why the cache is not really ding anything anymore.

It also shows that we lack performance tests.

PR #507, opened 2019-03-11
Feature/show inviter2

marcoconti83:
## What's new in this PR?

**Note**: this PR is replacing https://github.com/wireapp/wire-android-sync-engine/pull/506, that had to be rebased.

In the search results for team users, add the following results:
1. show a partner, if the current user was the one that invited the partner. This is relevant only if the invited person is a partner. If the invited person is not a partner, they it would have seen the invited person already anyway.
1. show the person that invited the current user. This is relevant only if the current user is a partner. If the current user is not a partner, then it would have seen the person that invited already anyway. 

### Solution

- For (1), I scan the team for people that are created by someone with my same user id.
- For (2), I fetch my user data from the team service using my team id, and check the `createdBy` field. @makingthematrix @CalumMcCall  I'm not sure if this is the best way to fetch my user data?

### Testing
The changes are covered by new unit tests. Additionally I added missing unit tests on the search. We were missing tests to check what happens if there is no query (empty query).

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
@@ -88,17 +88,16 @@ class UserSearchService(selfUserId:           UserId,
       case true if teamId.isDefined =>
         verbose(s"filterForPartner1 Q: $query, RES: ${searchResults.map(_.getDisplayName)}) with partner = true and teamId")
         for {
-          userThatInvitedMe <- usersStorage.get(selfUserId).map(_.flatMap(_.createdBy))
-          filteredUsers <- knownUsers.map(knownUsersIds => searchResults.filter(u =>
-            knownUsersIds.contains(u.id) || userThatInvitedMe.contains(u.id)
-          ))
+          Some(self)    <- userService.getSelfUser
+          filteredUsers <- knownUsers.map(knownUsersIds =>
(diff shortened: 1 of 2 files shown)


zmessaging/src/main/scala/com/waz/service/UserSearchService.scala
```diff
@@ -87,11 +87,18 @@ class UserSearchService(selfUserId:           UserId,
...
+          filteredUsers <- knownUsers.map(knownUsersIds => searchResults.filter(u =>
+            knownUsersIds.contains(u.id) || userThatInvitedMe.contains(u.id)
+          ))
+        } yield filteredUsers
       case false if teamId.isDefined =>
+        lazy val usersInvitedByMe = searchResults.filter(_.createdBy == Some(selfUserId)).map(_.id).toSet
         verbose(s"filterForPartner2 Q: $query, RES: ${searchResults.map(_.getDisplayName)}) with partner = false and teamId")
         knownUsers.map { knownUsersIds =>
           searchResults.filter { u =>
+            usersInvitedByMe.contains(u.id) ||
```
makingthematrix: This can be simplified. Notice that you filter `searchResults` twice - first for checking which users were invited by you, and then again, using what you got from the previous search. You can merge the two:
```
searchResults.filter { u =>
  u.createdBy.contains(selfUserId) ||
  knownUsersIds.contains(u.id) ||
  ...
}
```

PR #508, opened 2019-03-12
refactor: remove some redundancy in UserSearchServiceSpec

marcoconti83:
The previous code required the test author to repeat some information in functions call, that could actually be inferred from previous parameters. It was easy to make a mistake and pass in a different information the second time, leading to unexpected test results.

I removed the redundancy, requiring the test author to repeat the following information only once:
- query
- who is the self user

I also removed the need to pass the role for the user. Now it's inferred automatically from the user id.

### DIFF
--- a/zmessaging/src/test/scala/com/waz/service/UserSearchServiceSpec.scala
+++ b/zmessaging/src/test/scala/com/waz/service/UserSearchServiceSpec.scala
@@ -359,20 +359,16 @@ class UserSearchServiceSpec extends AndroidFreeSpec {
     /**
     * Helper class to keep track of mocked query
       */
-    class PreparedSearch(service: UserSearchService, query: String) {
+    case class PreparedSearch(inTeam: Boolean, selfId: UserId, query: String) {
 
       def perform() = {
-        this.service.search(query).map(_.local.map(_.id).toSet).head
+        val service = getService(this.inTeam, this.selfId)
+        service.search(query).map(_.local.map(_.id).toSet).head
       }
     }
 
     /**
     * Will mock all services, instantiate a UserSearchService to test, and store the query to expect
-      * @param query
-      * @param selfId
-      * @param conversationMembers
-      * @param connectedUsers
-      * @return
       */
     def prepareTestSearch(query: String,
                             selfId: UserId,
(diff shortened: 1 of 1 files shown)


zmessaging/src/test/scala/com/waz/service/UserSearchServiceSpec.scala
```diff
@@ -350,25 +348,44 @@ class UserSearchServiceSpec extends AndroidFreeSpec {
...
     }
   }
 
-  feature("search Inside the team") {
+  feature("search inside the team") {
+
+    /**
+    * Helper class to keep track of mocked query
+      */
+    class PreparedSearch(service: UserSearchService, query: String) {
```
makingthematrix: Since the service is also always prepared, I'd suggest to make `PreparedSearch` a case class with parameters: `inTeam: Boolean, selfId: UserId, query: String`, and create the service in the `perform` method.

zmessaging/src/test/scala/com/waz/service/UserSearchServiceSpec.scala
```diff
@@ -394,18 +411,27 @@ class UserSearchServiceSpec extends AndroidFreeSpec {
...
       }
+
+      // Set up user permissions according to role
+      val user = users(selfId)
+      userPrefs.setValue(UserPreferences.SelfPermissions, user.permissions._1)
+
+      // Return the service, configured for that user
+      val service = getService(true, selfId)
+
+      return new PreparedSearch(service, query)
```
makingthematrix: Please change `PreparedSearch` to a case class and then this line to just 
```
PreparedSearch(true, selfId, query)
```
(no `return new` needed)

makingthematrix:
There are also two places in the older part of the spec where you can use a bit more concise Scala syntax, but it does not change anything, so I decided only to write it here in a comment.
Instead of:
```
(userService.acceptedOrBlockedUsers _).expects().once().returning(Signal.const(expected.map(key => (key -> users(key))).toMap))
```
we can use:
```
(userService.acceptedOrBlockedUsers _).stubs().returning(Signal.const(users.filterKeys(connectedUsers.contains)))
```

and instead of:
```
Future.successful(ids.filter(i => conversationMembers.contains(i)).map(ConversationMemberData(_, convId)).toIndexedSeq)
```
we can use:
```
Future.successful(ids.intersect(conversationMembers).map(ConversationMemberData(_, convId)).toIndexedSeq)
```

PR #514, opened 2019-03-18
feat: add table to track notification latency

CalumMcCall:
### Issues

This PR adds the first table/repository required for notification tracking.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
+++ b/zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
@@ -290,7 +290,7 @@ object ZMessagingDB {
       db.execSQL("UPDATE KeyValues SET value = 'true' WHERE key = 'should_sync_teams'")
     },
     Migration(116, 117) { db =>
-      db.execSQL("CREATE TABLE FCMNotifications(_id TEXT PRIMARY KEY, received_at INTEGER, stage TEXT)")
+      db.execSQL(FCMNotificationsRepositoryDao.table.createSql)
     }
   )
 }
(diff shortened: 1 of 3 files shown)


zmessaging/src/main/scala/com/waz/service/push/FCMNotificationsRepository.scala
```diff
@@ -0,0 +1,66 @@
...
+case class FCMNotification(override val id: Uid,
+                           receivedAt:      Instant,
+                           stage:           String) extends Identifiable[Uid]
+
+object FCMNotification {
+
+  val Pushed = "pushed"
+  val Fetched = "fetched"
+  val StartOfPipeline = "StartOfPipeline"
+  val EndOfPipeline = "EndOfPipeline"
```
marcoconti83: Didn't we come to the conclusion that we don't need `EndOfPipeline` after all, as the record will be deleted when we get to that stage?

zmessaging/src/main/scala/com/waz/service/push/FCMNotificationsRepository.scala
```diff
@@ -0,0 +1,66 @@
...
+  */
+case class FCMNotification(override val id: Uid,
+                           receivedAt:      Instant,
+                           stage:           String) extends Identifiable[Uid]
+
+object FCMNotification {
+
+  val Pushed = "pushed"
+  val Fetched = "fetched"
+  val StartOfPipeline = "StartOfPipeline"
```
marcoconti83: For consistency, I would use the same verb tense as "pushed" and "fetched"

zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
```diff
@@ -287,6 +288,9 @@ object ZMessagingDB {
       db.execSQL("ALTER TABLE Users ADD COLUMN copy_permissions INTEGER DEFAULT 0")
       db.execSQL("ALTER TABLE Users ADD COLUMN created_by TEXT DEFAULT null")
       db.execSQL("UPDATE KeyValues SET value = 'true' WHERE key = 'should_sync_teams'")
+    },
+    Migration(116, 117) { db =>
```
yeryomenkom: We can avoid plain SQL here.

PR #516, opened 2019-03-19
Fix/logging global preference

johnxnguyen:
## What's new in this PR?

This PR contains some minor changes needed to use `LogsService` in the UI project.

### Issues

- the `LogsEnabled` preference is now a `GlobalPreference`. We don't currently support account independent logging, so it makes more sense to make the preference global.
- `LogsService` has been updated to handle only the global preference.
- the `LogsServiceImpl` singleton is owned by the global module, since it relies on `GlobalPreferences` which is also owned by the global module.
- the `clear` method has been implmented for `AndroidLogOutput`, so that logcat output will not persists if the user disables logs.


marcoconti83:
Last two commits look good to me

PR #515, opened 2019-03-18
refactor: Remove unused ReceivedPushData table

CalumMcCall:
### Issues

This commit removes the unused ReceivedPushData table, since it's now being replaced by FCMNotificationsRepository.

PR #517, opened 2019-03-19
Add stats table

CalumMcCall:
### Issues

Add the table of statistics which will be populated by notification tracking. This PR should be merged only after [ReceivedPushes PR](https://github.com/wireapp/wire-android-sync-engine/pull/515) has been merged.


zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
```diff
@@ -287,6 +285,16 @@ object ZMessagingDB {
       db.execSQL("ALTER TABLE Users ADD COLUMN copy_permissions INTEGER DEFAULT 0")
       db.execSQL("ALTER TABLE Users ADD COLUMN created_by TEXT DEFAULT null")
       db.execSQL("UPDATE KeyValues SET value = 'true' WHERE key = 'should_sync_teams'")
+    },
+    Migration(116, 117) { db =>
```
yeryomenkom: We can avoid plain SQL here.

PR #520, opened 2019-03-21
Add default rule for deriving LogShow instances.

yeryomenkom:
## What's new in this PR?

Default rule for deriving LogShow instances. We have a lot of cases where we use `LogShow.logShowWithHash`. I think that it safe enough to use it as a default type class instance.


marcoconti83:
So if not specialized, a type that we try to log will fallback to be hashed? Does it mean that all types are now considered safe for logging (hashed)?

marcoconti83:
@pwnsdx this PR is tweaking what is considered safe to log because it had become very cumbersome to explicitly allow every type.

yeryomenkom:
> So if not specialized, a type that we try to log will fallback to be hashed? Does it mean that all types are now considered safe for logging (hashed)?

Yes.

johnxnguyen:
@yeryomenkom I couldn't find any log show instances for primitive types. I thought it was intentional, but you may have forgotten to port them over to `LogShowInstancesSE`.

johnxnguyen:
@yeryomenkom oh no still one test failed.

PR #523, opened 2019-03-27
Write stages

CalumMcCall:
## What's new in this PR?

Write timestamps and stats for each stage

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/FCMNotificationStatsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/FCMNotificationStatsService.scala
@@ -44,17 +44,18 @@ class FCMNotificationStatsServiceImpl(fcmTimestamps: FCMNotificationsRepository,
 
   override def markNotificationsWithState(ids: Set[Uid], stage: String): Future[Unit] =  async {
     stage match {
-      case Pushed => await { Future.sequence(
-        ids.toSeq.map { fcmTimestamps.storeNotificationState(_, stage, Instant.now) })}
+      case Pushed => await {
+        Future.traverse(ids.toSeq)(p => fcmTimestamps.storeNotificationState(p, stage, Instant.now))
+      }
       case _ =>
-          val fcmIds = await(filterFCMNotifications(ids))
+          val fcmIds = await { filterFCMNotifications(ids) }
           verbose(l"""marking ${showString(fcmIds.size.toString)} notifications
                at stage ${showString(stage)} """)
-          await(calcAndStore(ids, stage))
+          await { calcAndStore(ids, stage) }
     }
   }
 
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/service/FCMNotificationStatsService.scala
```diff
@@ -17,46 +17,52 @@
...
 
-  override def markFCMNotificationsFetched(notifications: Seq[PushNotificationEncoded]): Future[Unit] = async {
-    val nots = notifications.map(_.id).toSet
-    val now = Instant.now
-    val fcmIds = await(filterFCMNotifications(nots))
-    val fut = fcmIds.map(id => storeNotificationState(id, FCMNotification.Fetched, now))
-    await(Future.sequence(fut))
+  override def markNotificationsWithState(ids: Set[Uid], stage: String): Future[Unit] =  async {
+    stage match {
+      case Pushed => await { Future.sequence(
```
yeryomenkom: Use `Future.traverse` in such case

zmessaging/src/main/scala/com/waz/service/FCMNotificationStatsService.scala
```diff
@@ -17,46 +17,52 @@
...
+        ids.toSeq.map { fcmTimestamps.storeNotificationState(_, stage, Instant.now) })}
+      case _ =>
+          val fcmIds = await(filterFCMNotifications(ids))
```
yeryomenkom: Use curly braces for `await` invocations.

zmessaging/src/main/scala/com/waz/service/FCMNotificationStatsService.scala
```diff
@@ -17,46 +17,52 @@
...
-  override def storeNotificationState(id: Uid, stage: String, timestamp: Instant): Future[Unit] =
-    async {
-      await { fcmTimestamps.storeNotificationState(id, stage, timestamp) }
-      val prevStage = await { fcmTimestamps.getPreviousStageTime(id, stage) }
-      if (prevStage.isDefined)
-        await { fcmStats.insertOrUpdate(getStageStats(stage, timestamp, prevStage.get)) }
-      if (stage == FCMNotification.FinishedPipeline) await { fcmTimestamps.deleteAllWithId(id) }
-    }
+  private def calcAndStore(ids: Set[Uid], stage: String): Future[Unit] = Future.sequence(ids.map { id =>
+    fcmTimestamps.getPreviousStageTime(id, stage).flatMap {
```
yeryomenkom: `Future.traverse`

zmessaging/src/main/scala/com/waz/service/FCMNotificationStatsService.scala
CalumMcCall: Ah, I didn't realize `traverse` is simply a parallel version of `sequence`, thanks for the hint!

PR #524, opened 2019-03-28
Fix: set preference default

johnxnguyen:
### Issues

In https://github.com/wireapp/wire-android-sync-engine/pull/522 @yeryomenkom added the possibility to set logging enabled for DEBUG builds only. However logs were still not being shown in such builds. 

### Causes

The `GlobalPreferences.LogsEnabled` always has a value (defaults to `false`), hence the `logsEnabledByDefault` flag is never considered.

### Solutions

Remove the `logsEnabledByDefault` flag and use the `DEBUG` flag as the preference default.


johnxnguyen:
build again

PR #525, opened 2019-03-28
Add trimming in case of bugs where notification stats aren't processed

CalumMcCall:
### Issues

If there are bugs in how we delete the timestamps of notifications, then we might end up with an ever-growing table of timestamps. To avoid this, a check is added when we write the final timestamp to delete the oldest row over a limit. This limit has been set to 1000 rows as this should be enough for even extreme situations.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/repository/FCMNotificationStatsRepository.scala
+++ b/zmessaging/src/main/scala/com/waz/repository/FCMNotificationStatsRepository.scala
@@ -46,7 +46,8 @@ trait FCMNotificationStatsRepository {
   def listAllStats(): Future[Vector[FCMNotificationStats]]
 }
 
-class FCMNotificationStatsRepositoryImpl(implicit db: Database)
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/repository/FCMNotificationStatsRepository.scala
```diff
@@ -65,7 +66,12 @@ class FCMNotificationStatsRepositoryImpl(implicit db: Database)
       })
       if (update.stage == FCMNotification.FinishedPipeline)
         FCMNotificationsDao.deleteEvery(FCMNotification.everyStage.map((stageTimestamp.id, _)))
-    }.map(_ => ())
+    }.andThen { case _ =>
+      //this op is not done as part of the transaction because it's not essential, it's just an
+      //extra check to ensure the FCM timestamps table doesn't grow overly long due to bugs
+      if (update.stage == FCMNotification.FinishedPipeline)
+        fcmTimestamps.trimExcessRows()
+    }(Threading.Background)
```
yeryomenkom: Request `ExecutionContext' at the class constructor.

PR #528, opened 2019-04-01
Feat/notification tracking

CalumMcCall:
### Issues

This PR introduces a new feature where we track how long it takes to process all notifications received over FCM. The results are persisted and displayed in the advanced options. These stats are visible on debug builds only, however they are always added to debug reports so users can report their results.


zmessaging/src/main/scala/com/waz/service/push/PushService.scala
```diff
@@ -165,13 +167,16 @@ class PushServiceImpl(selfUserId:           UserId,
...
     notificationStorage.getDecryptedRows().flatMap { rows =>
       verbose(l"Processing ${rows.size} rows")
-      if (rows.nonEmpty)
+      if (rows.nonEmpty) {
+        val ids = rows.map(_.pushId).toSet
         for {
+          _ <- fcmService.markNotificationsWithState(ids, StartedPipeline)
           _ <- pipeline(rows.flatMap(decodeRow))
           _ <- notificationStorage.removeRows(rows.map(_.index))
+          _ <- fcmService.markNotificationsWithState(ids, FinishedPipeline)
```
marcoconti83: Is this the right place to mark the end of pipeline? Why not after `processDecryptedRows()`?

CalumMcCall: Good question, so I did it this way because `processDecryptedRows` is recursive. This recursive call is often going to be the end of processing, as there will be no rows left. The `removeRows` call above this line marks the point where the notifications are marked as processed. We call recursively only in case additional notifications arrived whilst the pipeline was running.

marcoconti83: Thanks for clarifying

PR #531, opened 2019-04-09
Sync users from the backend

makingthematrix:
All that is needed for the "Email in the rich profile" feature in SE is that we should re-import user data from the backend. If this is done, we should get updated emails which then will be displayed in the rich profile.


makingthematrix:
build again

Issue #532, opened 2019-04-11
websocket connection unreliable


jschaul:
### Issue

(especially problematic on phones without play services, but likely affecting everyone)

On a phone without play services, disabling then re-enabling mobile internet breaks the websockets in such a way that they don't recover; i.e. messages will never arrive, until the user force quits the app.

Note that "normal" quitting of the app (by swiping it from recently used apps) does not re-create a websocket connection. 
Note that, apart from manually enabling/disabling mobile internet, the same "broken state" can be observed up to multiple times daily in normal usage scenarios, in a no-playservices scenario (possibly triggered by changing wifi networks or having good/bad mobile reception; but could be triggered otherwise, too)

### Causes

Unknown, possibly due to bugs in okhttp or the usage of that library.

### Solutions

- Upgrading okhttp may help as a first step. See changelog (3.9.1 and 3.10 seem most relevant regarding websockets): https://github.com/square/okhttp/blob/master/CHANGELOG.md
- possibly `pingInterval()` should be used.

### Testing

1. current android apk on a phone without play services: open the app, check messages arrive
2. disable mobile internet
3. enable mobile internet
4. send more messages - messages do not arrive anymore (the websocket never reconnects). 

(disabling wifi does not exhibit this behaviour; killing the websocket connection on a server also not - in these cases a websocket connection is correctly re-established)

PR #533, opened 2019-04-11
upgrade okhttp to newer version

jschaul:
## What's new in this PR?

~~Possibly fixes, or at least improves upon #532 - to be tested once I get an APK built with this change in. (I currently have a problem with sbt/scala)~~

There are newer versions of this dependency than 3.12.1, but they're not backwards compatible for android < 5.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -142,7 +142,7 @@ lazy val zmessaging = project
       compilerPlugin("com.github.ghik" %% "silencer-plugin"       % "0.6"),
 
       "org.scala-lang.modules"        %% "scala-async"           % "0.9.7",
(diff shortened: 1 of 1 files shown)


yeryomenkom:
@jschaul thank you for your proposal, but we need some time for investigations.

jschaul:
I built a custom APK and can confirm that this change alone **does not fix issue #532**. Still, upgrading this library might be good for all the other bugfixes included, so leaving this PR open.

Possibly, in addition to this dependency upgrade, `pingInterval()` could be used to detect (and refresh) a broken websocket connection.

jschaul:
made obsolete by #534

PR #534, opened 2019-04-16
Fix/fix websocket behaviour

jschaul:
### Issues

See #532 

### Causes

Any change to the network mode (i.e. disabling mobile internet, changing from 4G to wifi or vice-versa) is not noticed by the websocket code. This leads to the websocket **thinking** it's still alive while it is not.

### Solutions

Make use of okhttp's `pingInterval` feature - this sends a ping to the server, expects a pong, and if it does not, send a Closed event - which will then in turn re-establish a fresh websocket connection.

Currently I set the interval to 30 seconds (can be discussed what a good tradeoff with regards to battery consumption is)

### Testing

1. compile an APK
2. open the app, keep the app open, switch off and on mobile internet
3. before: no messages arrive. With this PR: after waiting up to 30 seconds, messages arrive again.

### Current behaviour:

```
2019-04-16T17:00:04.995Z-TID:1/V/DefaultNetworkModeService: updateNetworkMode: OFFLINE
2019-04-16T17:00:05Z-TID:1/W/AVS: NOT IMPLEMENTED: networkChanged

2019-04-16T17:00:12.012Z-TID:1/V/DefaultNetworkModeService: updateNetworkMode: _4G
2019-04-16T17:00:12.015Z-TID:1/W/AVS: NOT IMPLEMENTED: networkChanged
```

At this point, the websocket connection is effectively broken (no new messages arrive), but OkHttpWebSocketFactory and WSPushService did not notice this. There is no log entry showing that a `SocketEvent.Closed` has been sent. This will never recover on a phone without play services - not even after hours. This may recover on a phone with play services, but only if the app is put to the background or quit then re-opened.

### After this PR:

After a change to network mode:

```
2019-04-16T17:43:45.536Z-TID:1/V/DefaultNetworkModeService: updateNetworkMode: OFFLINE
2019-04-16T17:43:45.540Z-TID:1/W/AVS: NOT IMPLEMENTED: networkChanged

2019-04-16T17:43:53.734Z-TID:1/V/DefaultNetworkModeService: updateNetworkMode: _4G
2019-04-16T17:43:53.740Z-TID:1/W/AVS: NOT IMPLEMENTED: networkChanged

(...)
```

up to 30 seconds later we can see a "didn't receive pong", which then fails the websocket connection:

```
2019-04-16T17:44:16.164Z-TID:1641/W/OkHttpWebSocketFactory$: WebSocket connection has been failed.
java.net.SocketTimeoutException: sent ping but didn't receive pong within 30000ms (after 1 successful ping/pongs)
	at okhttp3.internal.ws.RealWebSocket.writePingFrame(RealWebSocket.java:546)
	at okhttp3.internal.ws.RealWebSocket$PingRunnable.run(RealWebSocket.java:530)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:428)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:278)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:273)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
	at java.lang.Thread.run(Thread.java:761)
(...)
```

which then enters the normal routine to restart the websocket connection, which is what we want.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -142,7 +142,7 @@ lazy val zmessaging = project
       compilerPlugin("com.github.ghik" %% "silencer-plugin"       % "0.6"),
 
       "org.scala-lang.modules"        %% "scala-async"           % "0.9.7",
-      "com.squareup.okhttp3"          %  "okhttp"                % "3.12.1",
+      "com.squareup.okhttp3"          %  "okhttp"                % "3.10.0",
       "com.googlecode.libphonenumber" %  "libphonenumber"        % "7.1.1", // 7.2.x breaks protobuf
       "com.wire"                      %  "cryptobox-android"     % "1.0.0",
       "com.wire"                      %  "generic-message-proto" % "1.23.0",
--- a/build.sbt
+++ b/build.sbt
@@ -171,6 +171,7 @@ lazy val zmessaging = project
       "org.robolectric"               %  "android-all"           % RobolectricVersion % Test,
       "junit"                         %  "junit"                 % "4.8.2"            % Test, //to override version included in robolectric
       "io.fabric8"                    %  "mockwebserver"         % "0.1.0"            % Test,
+      "com.squareup.okhttp3"          %  "mockwebserver"         % "3.10.0"           % Test, //should match okhttp version
       "org.apache.httpcomponents"     %  "httpclient"            % "4.5.3"            % Test
 //      "com.wire.cryptobox" % "cryptobox-jni" % cryptoboxVersion % Test % Native classifier "darwin-x86_64",
 //      "com.wire.cryptobox" % "cryptobox-jni" % cryptoboxVersion  % Test % Native classifier "linux-x86_64",
--- a/build.sbt
+++ b/build.sbt
@@ -142,7 +142,7 @@ lazy val zmessaging = project
       compilerPlugin("com.github.ghik" %% "silencer-plugin"       % "0.6"),
 
       "org.scala-lang.modules"        %% "scala-async"           % "0.9.7",
-      "com.squareup.okhttp3"          %  "okhttp"                % "3.9.0",
+      "com.squareup.okhttp3"          %  "okhttp"                % "3.12.1",
       "com.googlecode.libphonenumber" %  "libphonenumber"        % "7.1.1", // 7.2.x breaks protobuf
       "com.wire"                      %  "cryptobox-android"     % "1.0.0",
       "com.wire"                      %  "generic-message-proto" % "1.23.0",
--- a/zmessaging/src/main/scala/com/waz/znet2/WebSocketFactory.scala
+++ b/zmessaging/src/main/scala/com/waz/znet2/WebSocketFactory.scala
@@ -25,6 +25,7 @@ import com.waz.znet2.WebSocketFactory.SocketEvent
 import com.waz.znet2.http.{Body, Request}
 import okio.ByteString
 import org.json.JSONObject
+import java.util.concurrent.TimeUnit
 
 import scala.util.Try
 
@@ -56,7 +57,9 @@ object OkHttpWebSocketFactory extends WebSocketFactory with DerivedLogTag {
   import HttpClientOkHttpImpl.convertHttpRequest
 
   //TODO Should be created somewhere outside
-  private lazy val okHttpClient = new OkHttpClient()
+  private lazy val okHttpClient = new OkHttpClient.Builder()
+                                      .pingInterval(30000,  TimeUnit.MILLISECONDS)
+                                      .build();
 
   override def openWebSocket(request: Request[Body]): EventStream[SocketEvent] = {
     new EventStream[SocketEvent] {


marcoconti83:
Hi @jschaul ,
Thanks for this great PR, we have heard reports of related issues but we could not nail it down. 👍 

In this PR you are adding pings, but also bumping OkHTTP. Are the two things related, or can we split them into separate PR? In that case I would add the pings first, and test that one in internal, and only later try to bump OkHTTP

jschaul:
> In this PR you are adding pings, but also bumping OkHTTP. Are the two things related, or can we split them into separate PR?

Yes, they are related. In 3.10.0, using `pingInterval` checks to see if it receives a `pong` back. Previously, it did not, possibly not noticing that the connection broke. See the relevant changelog for okhttp [3.10](https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-3100):

> The pingInterval() feature now aggressively checks connectivity for web sockets and HTTP/2 connections.

> Previously if you configured a ping interval that would cause OkHttp to send pings, but it did not track whether the reply pongs were received. With this update OkHttp requires that every ping receive a response: if it does not the connection will be closed and the listener's onFailure() method will be called.

> (...)

> If you have a configured ping interval, you should confirm that it is long enough for a roundtrip from client to server. If your ping interval is too short, slow connections may be misinterpreted as failed connections. A ping interval of 30 seconds is reasonable for most use cases.

If you prefer, okhttp could be bumped to `3.10.0` first, if you have concerns that the changes between 3.10 and 3.12.1 require separate tests. Is there anything in the changelog that appears like it could lead to problems?

marcoconti83:
According to @yeryomenkom , he tried to build locally from your other PR where okhttp was bumped, and his client was misbehaving.

I will now try to build this PR locally and do some manual testing.

jschaul:
As I mentioned, for this functionality, `3.10.0` is sufficient. Would you like me to use that version in this PR in a first step? (A further bump to okhttp 1.12.1 could be done later)

marcoconti83:
I double-checked what the issues were with @yeryomenkom, the issues he was having were with compiling and testing, not with running.

I'm able to compile the client and on very simple manual smoke test it works as expected, however after the bump some of the tests on sync engine are failing, in particular:
- `com.waz.sync.client.AssetClient2Spec`
- `com.waz.znet2.OkHttpWebSocketSpec`
- `com.waz.znet2.HttpClientSpec`

Can you reproduce the same failures on your side? I will try to locally downgrade to 3.10 and see if the tests are still failing; if that's the case, we need to look into the specific reasons for the failures.

marcoconti83:
By downgrading to '3.10', I only get one reliably failing test: `com.waz.znet2.OkHttpWebSocketSpec`.

I'll try to figure out what's the issue with the test.

marcoconti83:
After further investigation, I discovered that our tests rely on a mock server from [fabric8](http://fabric8.io) that is tied to okhttp (`io.fabric8.`). From the website, the project seems to be not maintained anymore, and the latest version of this package only supports the okhttp 3.9 API, causing a run-time crash in the test due to API mismatch. 

We'll need some time to remove fabric8 from our tests.

marcoconti83:
I'm quite eager to get this in to address current issues in production builds, but I'm not sure what's the best way going forward.
 
`io.fabric8.webservermock` is [used here](https://github.com/wireapp/wire-android-sync-engine/blob/develop/zmessaging/src/test/scala/com/waz/znet2/OkHttpWebSocketSpec.scala#L41) only to mock a websocket server so that we can test our [`OkHttpWebSocketFactory`](https://github.com/wireapp/wire-android-sync-engine/blob/develop/zmessaging/src/main/scala/com/waz/znet2/WebSocketFactory.scala) implementation, which is basically a thin wrapper around okhttp that creates a connection and converts web socket events to the `SocketEvent` abstraction. 

@jschaul @yeryomenkom can you suggest a drop-in replacement for a websocket mock in Scala, or do you have a better suggestion on how to test the `OkHttpWebSocketFactory`?

jschaul:
> can you suggest a drop-in replacement for a websocket mock in Scala

https://github.com/square/okhttp/tree/master/mockwebserver maybe? Oddly enough, just adding the dependency in (see last commit, 90013990cedddd45b6699dc1dc7954daecab752a) results in the `com.waz.znet2.OkHttpWebSocketSpec` testcase no longer failing for me:

```
sbt "testOnly *OkHttpWebSocketSpec"
```

There are some other tests failing on this repo though, but I don't know if any of them are related, flaky, or due to my particular machine setup?

jschaul:
@marcoconti83 @CalumMcCall Is there anything more I can do here to help getting this PR merged?

marcoconti83:
I'm trying to fix the test by using the `OkHttpWebSocketSpec` by replacing `io.fabric8.webservermock`, which is deprecated, with the one you mentioned (`okhttp3.webservermock`). 

However their API are quite different and it seems like there's no easy way to inject events into the websocket. They (`okhttp`) do this in their tests [here](https://github.com/square/okhttp/blob/master/okhttp/src/test/java/okhttp3/internal/ws/WebSocketRecorder.java), but that testing infrastructure (`WebSocketRecorder`) is not exported.

I'm currently trying to port this approach to our codebase, but Scala is not my forte. If your Scala skills are hitching, you could try to port `WebSocketRecorder` yourself.

jschaul:
This PR is obsolete now, see #544 instead.

PR #535, opened 2019-04-18
Feature: custom backend

johnxnguyen:
## What's new in this PR?

This PR introduces the ability download backend configurations and to switch backends using that configuration.

`BackendConfig` has been modified to contain internal state for all properties that will change and accessor methods to read the current values. So far these are just the backend name and various endpoints.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/BackendConfig.scala
+++ b/zmessaging/src/main/scala/com/waz/service/BackendConfig.scala
@@ -22,10 +22,10 @@ import com.waz.utils.wrappers.URI
 import com.waz.znet.ServerTrust
 
 
-class BackendConfig(private var environment: String,
-                    private var baseUrl: URI,
-                    private var websocketUrl: URI,
-                    private var blacklistHost: URI,
+class BackendConfig(private var _environment: String,
+                    private var _baseUrl: URI,
+                    private var _websocketUrl: URI,
+                    private var _blacklistHost: URI,
                     val firebaseOptions: FirebaseOptions,
                     val pin: CertificatePin = ServerTrust.wirePin) {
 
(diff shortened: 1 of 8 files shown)


johnxnguyen:
build again

zmessaging/src/main/scala/com/waz/utils/PasswordValidator.scala
```diff
@@ -1,6 +1,6 @@
 /*
  * Wire
- * Copyright (C) 2019 Wire Swiss GmbH
+ * Copyright (C) 2016 Wire Swiss GmbH
```
marcoconti83: Weird jump back in time :)

zmessaging/src/main/scala/com/waz/service/BackendConfig.scala
```diff
@@ -17,11 +17,50 @@
...
+class BackendConfig(private var _environment: String,
+                    private var _baseUrl: URI,
+                    private var _websocketUrl: URI,
+                    private var _blacklistHost: URI,
+                    val firebaseOptions: FirebaseOptions,
+                    val pin: CertificatePin = ServerTrust.wirePin) {
+
+  val pushSenderId: String = firebaseOptions.pushSenderId
+
+  // We're using getter methods here to ensure after updating the config, any objects
```
marcoconti83: I think the comment is necessary in the context of the PR, but not needed in the code permanently. 

As long as whoever looks at this code understands the concept of getter/read-only variables, there is no need to explain why (and how) they are used. As the usage of the variables will change over time at the usage site, without anyone going back to this implementation to update the comment, the comment here will become less and less relevant/accurate.

I would move the comment to the commit/PR comments, where it would be useful to give context to the changes if anyone were to look at the commit history (blame) in some months from now.

zmessaging/src/main/scala/com/waz/service/BackendConfig.scala
```diff
@@ -17,11 +17,46 @@
...
+                    private var _blacklistHost: URI,
+                    val firebaseOptions: FirebaseOptions,
+                    val pin: CertificatePin = ServerTrust.wirePin) {
+
+  val pushSenderId: String = firebaseOptions.pushSenderId
+  
+  def environment: String = _environment
+  def baseUrl: URI = _baseUrl
+  def websocketUrl: URI = _websocketUrl
+  def blacklistHost: URI = _blacklistHost
```
johnxnguyen: We're using getter methods here to ensure after updating the config, any objects using it will always have up to date values. This is important when we switch backend configs.

johnxnguyen:
build again

PR #538, opened 2019-04-30
feat: Status Notifications

makingthematrix:
Depending on the availability status we restrict what notifications (if any) are pushed to UI to be displayed.
I had to add new requirements to the way `NotificationService` figures out what notifications to push to UI, and since the logic became a bit complex for one long expression, I decided to break it down in a series of i`f/else`s for better readability.
Also, I added unit tests, and refactored `NotificationServiceSpec`, putting code common to most unit tests in a private method.
Two new user preferences are going to be used in UI for displaying / not-displaying popups warning about the new functionality.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
@@ -134,36 +134,43 @@ class NotificationService(selfUserId:      UserId,
   })
 
   private def pushNotificationsToUi(): Future[Unit] = {
+    def shouldShowNotification(self: UserData,
+                               n: NotificationData,
+                               conv: ConversationData,
+                               notificationSourceVisible: Map[UserId, Set[ConvId]]): Boolean = {
+      // broken down for better readability
+      val appAllows =
+        n.user != self.id &&
+        conv.lastRead.isBefore(n.time) &&
+        conv.cleared.forall(_.isBefore(n.time)) &&
+        !notificationSourceVisible.get(self.id).exists(_.contains(n.conv))
+      val isReplyOrMention = n.isSelfMentioned || n.isReply
+      if (!appAllows) {
+        false
+      } else {
+        if (self.availability == Availability.Away) {
+          false
+        } else if (self.availability == Availability.Busy) {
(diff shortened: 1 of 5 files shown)


zmessaging/src/test/scala/com/waz/service/notifications/NotificationServiceSpec.scala
```diff
@@ -86,26 +133,54 @@ class NotificationServiceSpec extends AndroidFreeSpec with DerivedLogTag {
...
+      (uiController.onNotificationsChanged _).expects(account1Id, *).onCall { (_, nots) =>
+        nots.size shouldEqual 0
+        Future.successful({})
+      }
+
+      result(getService().messageNotificationEventsStage(rConvId, Vector(event)))
+    }
+
+    scenario("Push notifications to UI when the user is busy and the message is a reply/mention") {
+      val origMsg =
```
johnxnguyen: This only tests that the notification is presented when the user is busy and the message is a reply. Shouldn't we also do a test when the message is a reply?

zmessaging/src/main/scala/com/waz/service/push/NotificationService.scala
```diff
@@ -135,21 +136,39 @@ class NotificationService(selfUserId:      UserId,
...
-          conv.lastRead.isBefore(n.time) &&
-          conv.cleared.forall(_.isBefore(n.time)) &&
-          !notificationSourceVisible.get(selfUserId).exists(_.contains(n.conv))
-      }
-      _ = verbose(l"filteredShow: $filteredShow, seen: $seen")
-      _ <- uiController.onNotificationsChanged(selfUserId, filteredShow)
-      _ <- storage.insertAll(filteredShow.map(_.copy(hasBeenDisplayed = true)))
-      _ <- storage.removeAll(seen.map(_.id))
+      convs                     <- convs.getAll(toShow.map(_.conv).toSet).map(_.collect { case Some(c) => c.id -> c }.toMap)
+      (show, ignore)            =  toShow.partition { n =>
```
yeryomenkom: Move this huge condition into separate function please.

PR #541, opened 2019-05-08
fix: Adding a bitmask to the Availability status, to help handle statuses in UI

makingthematrix:
This is a part of the fix to https://wearezeta.atlassian.net/browse/AN-6200
We need to display a warning popup for each availability status, and then allow the user
to click "Never show that (one particular) warning again".
Instead of keeping a separate flag for each one I decided to keep one bitmask. This solution may also come in handy in the future, in other situations when we have to handle availability statuses.


makingthematrix:
build again

PR #543, opened 2019-05-09
fix: If the conversation is set to mentions & replies, a call should also go through

makingthematrix:
("mentions & replies" really means "mentions, replies & calls")

## What's new in this PR?

When we introduced mentions & replies, we added a possibility to filter notifications from a conversation, so the user was notified only when a message in the conversation was a reply or a mention. For some reason, probably connected to the fact that notifications about call work differently in iOS and Android, and we didn't catch that it would be a problem in Android, the filter ignored calls. This was caught only now, when the decision tree for notifications was updated because now notifications are also affected by the availability status. 

This PR changes the filter so that calls also go through if the conversation is set to "Mentions & Replies only". Basically, this mode means "Mentions, Replies & Calls".


makingthematrix:
build again

makingthematrix:
@johnxnguyen :
Generally yes, in Android, if requirements for a call are met, we display the call screen. There is a notification as well, but we always displayed both. The user may decide to dismiss the call screen, and then use the notification to join the call. But the app is always launched.

PR #545, opened 2019-05-10
fix: If the conversation is set to mentions & replies, a call should also go through

makingthematrix:
The SE part of https://github.com/wireapp/wire-android/pull/2126
Also, fixes the incosistency of reporting the conversation's mute status among platforms.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/MuteSet.scala
+++ b/zmessaging/src/main/scala/com/waz/model/MuteSet.scala
@@ -21,10 +21,8 @@ import MuteSet._
 import com.waz.log.LogShow.SafeToLog
 
 case class MuteSet(private val status: Set[MuteMask]) extends SafeToLog {
-  def reverse: MuteSet = MuteSet(MuteSet.AllMuted.status -- status)
-
   def toInt: Int =
-    (if (status.contains(StandardMuted)) 2 else 0) | (if (status.contains(MentionsMuted)) 1 else 0)
+    (if (status.contains(StandardMuted)) 1 else 0) | (if (status.contains(MentionsMuted)) 2 else 0)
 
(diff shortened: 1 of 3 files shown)


zmessaging/src/main/scala/com/waz/model/MuteSet.scala
```diff
@@ -39,23 +37,23 @@ object MuteSet {
...
+  // 2 -> `10` -> Only normal notifications are displayed (mentions are muted) -- not used
   // 3 -> `11` -> No notifications are displayed
-  def apply(status: Int): MuteSet =
-    MuteSet(
-      (if ((status & 1) != 0) Set[MuteMask](MentionsMuted) else Set.empty[MuteMask]) ++
-        (if ((status & 2) != 0) Set[MuteMask](StandardMuted) else Set.empty[MuteMask])
-    )
+  def apply(status: Int): MuteSet = MuteSet(status match {
+    case 1 | 2 => Set[MuteMask](StandardMuted) // fix for AN-6210, can be removed after some time (Maciek)
+    case 3 => Set[MuteMask](MentionsMuted, StandardMuted)
```
johnxnguyen: Why are we treating status 1 and 2 as identical? I thought we're not using status 2 at all.

makingthematrix: The mute status is stored in the conversation data in the db. Right now, if the user has set the mute status to "Mentions only", the stored integer for that is invalid. We need a way to read it and convert it correctly, so that the next time it will be updated. It could be also done in a migration, but I don't want to include a migration in this PR. I can start another for that.

johnxnguyen: Ah right 👍 If we intend to do a migration we should make a ticket about its so we don't forget.

PR #548, opened 2019-06-05
fix: Ignore markdown links

makingthematrix:
There should be no preview displayed for a markdown link. Unfortunately, we parse the message data in SE and here we prepare links to be displayed with previews, but only much later in UI we apply markdown. When the parsed link part of the message is displayed in UI we no longer know that originally it was a markdown link, and it's difficult to find it out at that point.
Instead, I decided to ignore markdown links during the parsing in SE. It's a bit hacky (now even if there
was no markdown handling in UI we would still not display the preview), but it's much easier and faster.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/MessageData.scala
+++ b/zmessaging/src/main/scala/com/waz/model/MessageData.scala
@@ -43,6 +43,7 @@ import org.threeten.bp.Instant.now
 
 import scala.collection.breakOut
 import scala.concurrent.duration._
+import scala.util.matching.Regex
 
 case class MessageData(override val id:   MessageId              = MessageId(),
                        convId:            ConvId                 = ConvId(),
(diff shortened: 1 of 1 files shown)


zmessaging/src/main/scala/com/waz/model/MessageData.scala
```diff
@@ -286,14 +283,18 @@ object MessageContent extends ((Message.Part.Type, String, Option[MediaAssetData
...
   val Empty = new MessageData(MessageId(""), ConvId(""), Message.Type.UNKNOWN, UserId(""))
   val Deleted = new MessageData(MessageId(""), ConvId(""), Message.Type.UNKNOWN, UserId(""), state = Message.Status.DELETED)
   val isUserContent = Set(TEXT, TEXT_EMOJI_ONLY, ASSET, ANY_ASSET, VIDEO_ASSET, AUDIO_ASSET, RICH_MEDIA, LOCATION, KNOCK)
 
   val EphemeralMessageTypes = Set(TEXT, TEXT_EMOJI_ONLY, KNOCK, ASSET, ANY_ASSET, VIDEO_ASSET, AUDIO_ASSET, RICH_MEDIA, LOCATION)
 
+  // A markdown link looks like that: [place for the text](here.goes.the.link)
+  // Links of this type will be handled by our Markdown library, we should ignore them here.
+  val markdownLinkPattern = """\[.*?\]\((.*?)\)""".r
+
```
johnxnguyen: This pattern only weakly matches the inline link. It means that we could think we find a markdown link when there isn't one, and therefore not render a link preview, when we should.

Also, this pattern unfortunately doesn't handle reference links at all, meaning it wouldn't detect a markdown link when there is one, and so it won't skip the link preview, when it should.

But this is the problem with using regex to parse markdown. You can see the spec [here](https://spec.commonmark.org/0.29/#links)

PR #554, opened 2019-06-11
fix: read receipts not updating

CalumMcCall:
### Issues

This commit adds missing support for updating local read receipt setting when it's changed on another client. This fixes a [related issue](https://wearezeta.atlassian.net/browse/AN-6110) where the popup notifying of the change was not displaying.

### Causes

I think this small detail was simply left out during implementation of this feature

### Solutions

The missing event type decoding has been implemented.

### Testing

Manual testing has been done. The option was toggled using a web client whilst the correct behaviour was observed on the device.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/Event.scala
+++ b/zmessaging/src/main/scala/com/waz/model/Event.scala
@@ -211,6 +211,7 @@ object Event {
         case "user.client-add" => OtrClientAddEvent(OtrClient.ClientsResponse.client(js.getJSONObject("client")))
         case "user.client-remove" => OtrClientRemoveEvent(decodeId[ClientId]('id)(js.getJSONObject("client"), implicitly))
         case "user.properties-set" => PropertyEvent.Decoder(js)
+        case "user.properties-delete" => PropertyEvent.Decoder(js)
         case _ =>
           error(l"unhandled event: $js")
           UnknownEvent(js)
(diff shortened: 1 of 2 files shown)


CalumMcCall:
build again

marcoconti83:
@CalumMcCall  is there any unit test already on this that we can extend to cover this case? If not, can we add some? Thanks!

marcoconti83:
Test looks good, thanks!

marcoconti83:
build again

CalumMcCall:
build again

CalumMcCall:
build again

zmessaging/src/main/scala/com/waz/model/Event.scala
```diff
@@ -410,7 +411,11 @@ object PropertyEvent {
     override def apply(implicit js: JSONObject): PropertyEvent = {
       import PropertyKey._
       decodePropertyKey('key) match {
-        case ReadReceiptsEnabled => ReadReceiptEnabledPropertyEvent('value)
+        case ReadReceiptsEnabled => decodeString('type) match {
+          case "user.properties-set" => ReadReceiptEnabledPropertyEvent(1)
```
marcoconti83: Is a constant `1` always the good solution? It was `value` before, maybe value include `1` and `0` and we should parse it instead of assuming `1`. Just speculating.

CalumMcCall: Yeah it does look a little strange, but if we receive a 'set' message type, then it's always 1, and vice versa. So the `'value` field is redundant. The value should only ever be binary, so I think this is safe, and also clearer. I'm open to changing it though if you prefer.

marcoconti83: I would parse the value in case of `set`, just in case. @typfel @bennyn how's the implementation on other platforms?

The fact that is always `0` when `delete` it's probably because `0` is the default? (it makes no sense to specify the value of something that you deleted).

typfel: On iOS we try convert the value to a number and treat anything > 0 as true

https://github.com/wireapp/wire-ios-request-strategy/blob/47b9003001a86341719b8355e1427e189ac3f59a/Sources/Request%20Strategies/User/UserPropertyRequestStrategy.swift#L91

CalumMcCall: We also treat any value > 0 as true, we have a unit test checking this.

PR #555, opened 2019-06-12
refactor: uncomment out eventspec tests

CalumMcCall:
### Issues

There were some tests that were disabled pending fixes to base64, which was completed a while ago, so we should re-enable or remove them. [See here](https://wearezeta.atlassian.net/browse/AN-6248)

### Solutions

The calling test is no longer relevant, since we decode call events from a GenericMessage instead inside `OtrServiceImpl.scala`, the other test was re-enabled.


CalumMcCall:
build again

CalumMcCall:
build again

PR #556, opened 2019-06-12
fix: LRUFileCacheSpec failing due to directory name conflicts

CalumMcCall:
- We create test directories for certain types of tests, such as
   LRUFileCacheSpec, but we use milliseconds to ensure uniqueness
   of the directory name, which wasn't always the case as the method
   is called frequently. This commit adds a counter to ensure dir names
   are unique.
 - This addresses AN-6249, thus also fixing the flaky LRUCache tests

### DIFF
--- a/zmessaging/src/test/scala/com/waz/cache2/LruFileCacheSpec.scala
+++ b/zmessaging/src/test/scala/com/waz/cache2/LruFileCacheSpec.scala
@@ -113,7 +113,7 @@ class LruFileCacheSpec extends ZIntegrationSpec {
       }
     }
 
-    ignore("Putting file directly in cache should trigger cache cleanup if it is needed") {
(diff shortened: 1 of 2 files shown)


zmessaging/src/test/scala/com/waz/cache2/LruFileCacheSpec.scala
```diff
@@ -113,7 +113,7 @@ class LruFileCacheSpec extends ZIntegrationSpec {
       }
     }
 
-    scenario("Putting file directly in cache should trigger cache cleanup if it is needed") {
+    ignore("Putting file directly in cache should trigger cache cleanup if it is needed") {
```
marcoconti83: Why is this test disabled?

CalumMcCall: Ah good catch, fixed!

PR #557, opened 2019-06-14
Fix: team sync

johnxnguyen:
### Issues

The app would sometimes get stuck in a sync loop. When syncing team data, the response from the backend would fail to decode. This caused the team sync to fail, but the request would continuously retry.

### Causes

The Circe json parser would fail to decode the response object from the backend. We use auto derivation to encode/decode json objects. In order for it to work, we must provide a data class the exactly models the json response object (though, only for the properties that we care about). 

In this specific case, the `TeamData` object is being used to model a response, however it contains inconsistent property names and data types for those properties. 

### Solutions

Provide a `TeamDataResponse` object to specifically model the response. Then convert the response object to a `TeamData` instance once decoding completes.

### Testing

A single unit test to decode the json response from the backend.

## Notes

The bigger issue here is that `TeamData` is a domain object and shouldn't have anything to do with backend responses. Client objects communicating with the backend should have their own response models, which should then be converted to the corresponding domain object. Enforcing the separation will help to avoid such critical bugs in the future, as changes to the domain objects would not affect how responses from the backend are decoded.

## Bonus

As a bonus, this also fixes a bug where the team icon is not displayed.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
+++ b/zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
@@ -60,11 +60,17 @@ class TeamsClientImpl(implicit
   }
 
   override def getTeamData(id: TeamId): ErrorOrResponse[TeamData] = {
-
     Request.Get(relativePath = teamPath(id))
-      .withResultType[TeamData]
+      .withResultType[TeamDataResponse]
       .withErrorType[ErrorResponse]
-      .executeSafe
+      .executeSafe { response =>
+        TeamData(
+          response.id,
+          response.name,
+          response.creator,
+          Some(response.icon),
+          response.icon_key)
+      }
   }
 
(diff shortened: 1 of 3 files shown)


zmessaging/src/main/scala/com/waz/sync/client/TeamsClient.scala
```diff
@@ -60,11 +60,17 @@ class TeamsClientImpl(implicit
...
-      .withResultType[TeamData]
+      .withResultType[TeamDataResponse]
       .withErrorType[ErrorResponse]
-      .executeSafe
+      .executeSafe { response =>
+        TeamData(
+          response.id,
+          response.name,
+          response.creator,
+          Some(response.icon),
```
CalumMcCall: Can icon ever be null? If so, then shouldn't this be `Option(response.icon)`?

johnxnguyen: In the response it isn't optional, but for `TeamData` it is.

PR #564, opened 2019-06-20
fix: Markdown links should not generate previews also on the sender's side.

makingthematrix:
The previous fix of markdown links worked only on the receiver's side.


makingthematrix:
build again

PR #565, opened 2019-06-24
Fix: Set lastEventTime to now when creating a conversation

johnxnguyen:
### Issues

A newly created conversation does not always appear at the top of the conversation lis.

### Causes

Conversations the list are sorted by their `lastEventTime` property. When a conversation is created, this value is set to `RemoteInstant.Epoch`. Typically a new conversation will have an event which will update it to the present, but this isn't guaranteed to happen (such as when there is no network connection).

### Solutions

Set the `lastEventTime` property to the current time. This makes sense, since we can think of the creation as an event.

**Updated** I reverted the previous "solution" and instead updated the last event time when we add the conversation created message locally.

### Testing

Manually tested when creating a conversation with no network connection.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsContentUpdater.scala
+++ b/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsContentUpdater.scala
@@ -29,6 +29,7 @@ import com.waz.service.tracking.TrackingService
 import com.waz.sync.SyncServiceHandle
 import com.waz.threading.{CancellableFuture, SerialDispatchQueue}
 import com.waz.utils._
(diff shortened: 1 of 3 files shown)


CalumMcCall:
This makes sense, but don't we always add start messages to conversations when we create them? Shouldn't that be updating the `lastEventTime`?

johnxnguyen:
@CalumMcCall typically yes, but it doesn't seem to be the case when we're offline.

makingthematrix:
> This makes sense, but don't we always add start messages to conversations when we create them? Shouldn't that be updating the `lastEventTime`?

I think it should, but it doesn't. Let's talk tomorrow.

PR #567, opened 2019-06-26
fix: Self-replies don't show up on Android

makingthematrix:
## What's new in this PR?

A fix to a regression bug.
fixes: https://wearezeta.atlassian.net/browse/AN-6279

### Issues

When a user from another platform sent a reply to their own message, Android displayed an error message instead. 

### Causes

The bug was regression. It occured after the new assets were merged. In changes to `MessageEventProcessor` a call to the `checkReplyHashes` method was deleted. The call has to be there, because we need to recreate locally the hash of the quoted message. If not, the hashes may differ and the quoted message will be treated as invalid (look to the specs of of the replies feature for details).

### Solutions

The fix simply adds back the call to `checkReplyHashes`.

### Testing

Preconditions: User1 is on Android, User2 is on webapp, connected
1) Open 1:1 with User2
2) User 2 sends a message (text, audio etc)
3) User 2 replies on his own message/User1 replies on User2 messages

### Notes

There's only one new line in the whole file. The rest is refactoring.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
+++ b/zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
@@ -17,17 +17,12 @@
  */
 package com.waz.service.messages
 
-
 import com.waz.api.{Message, Verification}
 import com.waz.content.MessagesStorage
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.log.LogSE._
-import com.waz.model.AssetMetaData.Image.Tag.{Medium, Preview}
-import com.waz.model.AssetStatus.{UploadCancelled, UploadFailed}
-import com.waz.model.GenericContent.{Asset, Calling, Cleared, DeliveryReceipt, Ephemeral, ImageAsset, Knock, LastRead, LinkPreview, Location, MsgDeleted, MsgEdit, MsgRecall, Reaction, Text}
-import com.waz.model.{GenericContent, _}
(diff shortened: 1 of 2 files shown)


zmessaging/src/main/scala/com/waz/service/messages/MessageEventProcessor.scala
```diff
@@ -65,72 +60,73 @@ class MessageEventProcessor(selfUserId:          UserId,
...
-        case _                          => Future.successful(None)
-      }
-
-    } yield EventAndLocalData(event, message, asset)
-
-  def localDataForEvents(events: Seq[MessageEvent]): Future[Seq[EventAndLocalData]] =
-    Future.traverse(events)(localDataForEvent)
-
   private[service] def processEvents(conv: ConversationData, isGroup: Boolean, events: Seq[MessageEvent]): Future[Set[MessageData]] = {
     verbose(l"processEvents: ${conv.id} isGroup:$isGroup ${events.map(_.from)}")
```
johnxnguyen: This method is 68 lines long. It would be nice to refactor it to make it more readable.

makingthematrix: I can move out the two nested methods, but they are only used in the following for-comprehension, so it seems justified to have them nested.
Other than that, every line in the for-comprehension is needed, so I don't see much room for refactoring... any ideas?

johnxnguyen: Personally I find nested methods distracting, if I want to know more about a method invocation I will look up the definition when I want to and in context.

The for loop could be refactored by splitting out groups of statements into their own function / method, with a descriptive name. The aim would be to make it read like a sequence of sentences, rather than trying to understand the minute details of each statement.

These are just suggestions of course :)

PR #569, opened 2019-06-30
fix: youtube preview images

CalumMcCall:
### Issues

 - This commit adds the ability to specify an HTTP proxy at compile time
 - [AN-6028](https://wearezeta.atlassian.net/browse/AN-6028)
 - It was tested manually using tinyproxy locally

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
@@ -163,7 +163,7 @@ class ConversationsUiServiceImpl(selfUserId:        UserId,
     val messageId = MessageId()
     for {
       retention  <- messages.retentionPolicy2ById(convId)
-      rr <- readReceiptSettings(convId)
+      rr         <- readReceiptSettings(convId)
       rawAsset   <- assets.createAndSaveUploadAsset(content, AES_CBC_Encryption.random, public = false, retention, Some(messageId))
       message    <- messages.addAssetMessage(convId, messageId, rawAsset, rr, exp)
       _          <- updateLastRead(message)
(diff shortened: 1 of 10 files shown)


CalumMcCall:
Ah yeah that's an excellent question. So UploadAssetIds are not simply UUIDs, if you print them out you'll see a couple numbers prefixed to the UUID, which AFAIK isn't part of the standard for UUIDs. However, the fact they aren't interchangeable is implied by using different types. I just mean that potentially it affects encoding/decoding. However on reflection now they are all just treated as strings anyway. So the whole point is moot. Sorry for the confusion. It's just something I noticed whilst tested that surprised me(that the ids when printed out aren't all just UUIDS).

johnxnguyen:
Thanks for the clarification 😄

marcoconti83:
@CalumMcCall @johnxnguyen to clarify, the backend assed ID is a UUID prefixed with a bucket identifier, it reflects how they will be stored in the backend.

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsUiService.scala
```diff
@@ -162,15 +162,14 @@ class ConversationsUiServiceImpl(selfUserId:        UserId,
...
-      retention    <- messages.retentionPolicy2(conversation)
-      rr           <- readReceiptSettings(convId)
-      rawAsset     <- assets.createAndSaveUploadAsset(content, AES_CBC_Encryption.random, public = false, retention, Some(messageId))
-      message      <- messages.addAssetMessage(convId, messageId, rawAsset, rr, exp)
-      _            <- updateLastRead(message)
-      _            <- Future.successful(tracking.assetContribution(AssetId(rawAsset.id.str), selfUserId)) //TODO Maybe we can track raw assets contribution separately?
-      shouldSend   <- checkSize(convId, rawAsset, message, confirmation)
-      _            <- if (shouldSend) sync.postMessage(message.id, convId, message.editTime) else Future.successful(())
+      retention  <- messages.retentionPolicy2ById(convId)
+      rr <- readReceiptSettings(convId)
```
johnxnguyen: Could you indent this to make it inline with the other lines?

PR #570, opened 2019-07-03
feat: add http proxy support

CalumMcCall:
### Issues

 - This commit adds the ability to specify an HTTP proxy at compile time
 - AN-6291
 - It was tested manually using tinyproxy locally

### DIFF
--- a/zmessaging/src/main/scala/com/waz/znet2/WebSocketFactory.scala
+++ b/zmessaging/src/main/scala/com/waz/znet2/WebSocketFactory.scala
@@ -59,16 +59,15 @@ class OkHttpWebSocketFactory(proxy: Option[Proxy]) extends WebSocketFactory with
   import HttpClientOkHttpImpl.convertHttpRequest
 
   //TODO Should be created somewhere outside
-  private lazy val okHttpClient = if(proxy.isDefined) {
-    new OkHttpClient.Builder()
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/znet2/WebSocketFactory.scala
```diff
@@ -57,9 +59,16 @@ object OkHttpWebSocketFactory extends WebSocketFactory with DerivedLogTag {
   import HttpClientOkHttpImpl.convertHttpRequest
 
   //TODO Should be created somewhere outside
-  private lazy val okHttpClient = new OkHttpClient.Builder()
-                                      .pingInterval(30000,  TimeUnit.MILLISECONDS)
-                                      .build();
+  private lazy val okHttpClient = if(proxy.isDefined) {
+    new OkHttpClient.Builder()
+      .pingInterval(30000,  TimeUnit.MILLISECONDS)
```
marcoconti83: A comment on the code style: I would reduce code duplication and magic numbers by setting the `pingInterval` only once, e.g.:

```scala
private lazy val builder = new OkHttpClient.Builder().pingInterval(30000, TimeUnit.MILLISECONDS)
private lazy val okHttpClient = if(proxy.isDefined) {
   builder.proxy(proxy.get).build()
} else {
   builder.build()
}
```

... or something similar

PR #572, opened 2019-07-10
fix: Don't fail on retrieving asset details

makingthematrix:
This is the SE part of the fix.

When we send an asset message, first we try to get some details about it. The data in the `AssetDetails` entity depends on the asset's type: it's different for video, audio, image, or an unrecognized file. In the old version, we tried to get the data based on the Mime type which
was provided. When the mime was wrong, we were not able to get the data and the process failed.
Now, the process tries to find out what the mime is by itself. Initially we guess it based on the extension of the filename, but if we fail to retrieve the data for that mime type, we either assume the asset is an unrecognized file, or - in the case of video - we check first if the asset isn't an audio file (MP4 is an extension for both video or audio). Only if we fail to retrieve audio data, we assume it's an
unrecognized file. In all cases, we don't crash the process.


makingthematrix:
build again

PR #573, opened 2019-07-16
Fix: google map previews

johnxnguyen:
## What's new in this PR?

After the assets refactoring, the code used to load map previews died. This PR repurposes this code to help facilitate fetching map previews.

The map previews can now be loaded using Glide by simply providing a url. As such, we no longer have to return an `AssetData`, rather a `URI` now suffices.

## Edit

We couldn't make the direct request to google due to privacy issues, so I've reverted back to using the proxy. As a result, I've changed `GoogleMapsClient` to make the request to the backend as we do with other clients. `GoogleMapsMediaService` is responsible for invoking the request, and returns an `InputStream` containing the image data. This stream will be used in the UI to load via _Glide_.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/sync/client/GoogleMapsClient.scala
+++ b/zmessaging/src/main/scala/com/waz/sync/client/GoogleMapsClient.scala
@@ -20,13 +20,14 @@ package com.waz.sync.client
 import java.net.URLEncoder
 
 import com.waz.service.media.RichMediaContentParser.GoogleMapsLocation
+import com.waz.utils.wrappers.URI
 
 object GoogleMapsClient {
-  val StaticMapsPathBase = "/proxy/googlemaps/api/staticmap"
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/sync/client/GoogleMapsClient.scala
```diff
@@ -20,13 +20,14 @@ package com.waz.sync.client
 import java.net.URLEncoder
 
 import com.waz.service.media.RichMediaContentParser.GoogleMapsLocation
+import com.waz.utils.wrappers.URI
 
 object GoogleMapsClient {
-  val StaticMapsPathBase = "/proxy/googlemaps/api/staticmap"
+  val StaticMapsPathBase = "https://maps.googleapis.com/maps/api/staticmap"
```
marcoconti83: We have a backend proxy here to avoid exposind the device IP and search query to Google. We should keep using the proxy.

PR #574, opened 2019-07-16
Feat/64 bit builds

CalumMcCall:
### Issues

Wire has to support 64-bit by [August 1st](https://developer.android.com/distribute/best-practices/develop/64-bit). This PR changes the SE build to support 64-bit, it also uses a newer version(yet to be merged and published) of cryptobox-jni.

### Testing

This build has been verified to contain 64-bit versions of all libs by analysing it using Build->Analyze APK in Android Studio.


CalumMcCall:
build again

marcoconti83:
This is now obsolete, has been addressed otherwise

PR #576, opened 2019-07-17
fix: accept websocket uri without path

CalumMcCall:
What's new in this PR?

### Issues

iOS clients append the "/await" path to websocket uris automatically. On Android we need to add them ourselves. This was causing issues with using one config for both Android/iOS, so this PR changes the Android client to be able to accept both kinds of URIs, and append the required path when needed.

Fixes [AN-6312](https://wearezeta.atlassian.net/browse/AN-6312)


CalumMcCall:
build again

PR #577, opened 2019-07-17
fix: clearing conversations now clears failed msgs

CalumMcCall:
### Issues

 - Fixes [AN-6293](https://wearezeta.atlassian.net/browse/AN-6293)
 - This PR introduces an additional step to delete unsent(failed, pending) messages when clearing a conversation.

### Causes

Since (failed, pending) messages haven't been synced yet, they weren't being cleared, as the `lastEventTime` is used to set the cleared time, and this is only updated after a given message has been synced with the backend.

### Solutions

Since changing `lastEventTime` would result in other clients potentially losing messages, it was decided to simply delete all unsent messages when clearing, since this most closely matches what the user would expect

### Testing

It was tested manually, and clearing the conversation now deletes failed messages as expected

### DIFF
--- a/zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
+++ b/zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
@@ -35,6 +35,7 @@ import com.waz.threading.{CancellableFuture, SerialDispatchQueue}
 import com.waz.utils.TrimmingLruCache.Fixed
 import com.waz.utils._
 import com.waz.utils.events.{EventStream, Signal, SourceStream}
+import com.waz.utils.wrappers.DB
 
 import scala.collection.immutable.Set
 import scala.collection._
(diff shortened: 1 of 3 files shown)


zmessaging/src/main/scala/com/waz/content/MessagesStorage.scala
```diff
@@ -298,6 +300,16 @@ class MessagesStorageImpl(context:     Context,
...
     }
   }
+
+  private def clearUnsentMsgs(convId: ConvId)(implicit storage: DB): Unit = {
+    val unsentMessages = MessageDataDao.listUnsentMsgs(convId)
+    MessageDataDao.iterating(unsentMessages).foreach(_.foreach(clearUnsentMsg))
+  }
+
+  private def clearUnsentMsg(data: MessageData)(implicit storage: DB): Unit = {
+    MessageDataDao.delete(data.id)
```
johnxnguyen: Is this clearing an unsent message, or just a message?

CalumMcCall: `delete` removes any message with the given `MessageId`. Here these messages are only ever unsent messages.

johnxnguyen: Then it would be better to name this method `clearMsg` since the method alone has nothing to do with unsent messages.

CalumMcCall:
I simplified it more, turns out there was a `delete` method I overlooked. I also renamed the method to something better

PR #580, opened 2019-07-25
feat: add pref to hide screen content

CalumMcCall:
### Issues

 - [AN-6323](https://wearezeta.atlassian.net/browse/AN-6323)
 - This PR introduces a new user pref to track whether to hide screen
content in the task switcher
 - The method used in fact blocks all screenshots, including the ones
taken for the task switcher


CalumMcCall:
build again

